PR #72, opened 2025-07-09
create distinct colors per object

mrloureed:
image

https://viam.atlassian.net/browse/APP-8728

### DIFF
--- a/src/lib/color.ts
+++ b/src/lib/color.ts
@@ -72,12 +72,68 @@ export const colors = {
 		selected: oklchToHex(twColors.amber['900']),
 		default: oklchToHex(twColors.amber['500']),
 	},
+	base: {
+		selected: oklchToHex(twColors.slate['900']),
+		default: oklchToHex(twColors.slate['500']),
+	},
+	board: {
+		selected: oklchToHex(twColors.emerald['900']),
+		default: oklchToHex(twColors.emerald['500']),
+	},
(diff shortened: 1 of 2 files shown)


src/lib/components/Frame.svelte
```diff
@@ -20,13 +20,21 @@
 
 	const selected = useSelected()
 	const events = useObjectEvents(() => uuid)
+
+	const colorGroup = getColorGroup(rest.name)
```
micheal-parks, line 24:
```suggestion
	const colorGroup = $derived(getColorGroup(rest.name))
```
since name can change

src/lib/color.ts
```diff
@@ -72,12 +72,81 @@ export const colors = {
...
+	if (type === 'unnamed geometry') type = 'unnamed_geometry'
+	const group = colors[type as keyof typeof colors]
+	if (group && typeof group === 'object' && 'selected' in group && 'default' in group) {
+		return group
+	}
+	return {
+		selected: colors.selected,
+		default: colors.default,
+	}
+}
```
micheal-parks, line 152: using name here is going to cause issues since people can name resources anything - we'll need to change `useFrames` so that the results have the resource type attached. I can work on that

micheal-parks:
Closing, changes pulled into https://github.com/viam-labs/motion-tools/pull/73

PR #98, opened 2025-08-15
Improve pointer miss and custom geometry logic

micheal-parks:


src/lib/components/PointerMissBox.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
```
micheal-parks, line 1: Previously, deselecting the current selected item happened by attaching a `pointermiss` event handler to each object. This wasn't always reliable due to event ordering and wasn't compatible with TransformControls, which uses its own raycaster instead of Threlte's interactivity plugin.

I found it more reliable to add a big invisible box surrounding the scene to catch "deselect" clicks.

src/lib/components/Selected.svelte
```diff
@@ -8,41 +8,57 @@
...
+	})
+	const clone = $derived.by(() => {
+		if (selected.current?.metadata.batched) {
+			return
+		}
+
+		return object3d?.clone(false)
+	})
+
+	const { start, stop } = useTask(
```
micheal-parks, line 26: There was a bug where bounding boxes on the current item would not update in some circumstances. This just makes updates reliable by putting them into the frameloop when something is selected.

PR #106, opened 2025-09-05
Have consecutive calls to `DrawFrameSystem` attempt to keep the same geometry label -> color mapping.

dgottlieb:


micheal-parks (approved):
Thank you!

PR #109, opened 2025-09-05
Allow updating geometries via draw commands

micheal-parks:
FYI @dgottlieb thanks for reporting!

draw


client/client/client.go
```diff
@@ -164,6 +168,10 @@ func DrawGeometry(geometry spatialmath.Geometry, color string) error {
 
 // DrawGeometries draws a list of geometries in the visualizer.
 //
+// Labels must be unique within a world. Calling DrawGeometry with labels that
```
DTCurrie, line 171:
```suggestion
// Labels must be unique within a world. Calling DrawGeometries with labels that
```

PR #112, opened 2025-09-09
Improve world state service integration

DTCurrie:
I don't love this design, and you can see I took a few shortcuts, but the point was to get something that renders, so we have something to work with. I tested with a service animating at 30fps, and performance seemed pretty solid.


src/lib/hooks/useWorldState.svelte.ts
```diff
@@ -1,43 +1,76 @@
...
 
-export const useWorldStates = () => {
+const key = Symbol('world-state-context')
+
+interface Context {
+	names: ResourceName[]
+	current: Record<string, ReturnType<typeof createWorldState>>
+}
+
+const worker = new Worker(new URL('../workers/worldStateWorker.ts', import.meta.url), {
```
micheal-parks, line 26: Should we use vite's convienience stuff for web workers? I know that exporting libraries with them can get hairy: https://v3.vitejs.dev/guide/features.html#import-with-query-suffixes

micheal-parks (approved):

src/lib/WorldObject.svelte.ts
```diff
@@ -44,14 +54,62 @@ export class WorldObject<T extends Geometries = Geometries> {
...
+
+	for (const [k, v] of Object.entries(metadata)) {
+		const unwrappedValue = unwrapValue(v)
+
+		// TODO: Remove special case and add better handling for metadata
+		if (k === 'color' && unwrappedValue && typeof unwrappedValue === 'object') {
+			const { r, g, b } = unwrappedValue as RGB
+			json[k] = new Color().setRGB(r / 255, g / 255, b / 255)
+		} else {
+			json = { ...json, [k]: unwrappedValue }
```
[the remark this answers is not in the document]
DTCurrie: Yeah, all of this ugliness is because the TS-SDK uses the `PlainMessage` wrapper for most types, and that recursively removes method fields from types. We could have just done `metadata.toJson()` and be done with all of this but alas. I think I am going to rewrite this to just put in some TS-ignores for now and remove this. But as I mentioned in our convo, I think we just need to write a better metadata parser anyway since we have typed out some specific expectations. We may want to document those expectations as well for service module authors.

src/lib/hooks/useWorldState.svelte.ts
DTCurrie: See the [above section](https://v3.vitejs.dev/guide/features.html#import-with-constructors): 

> A web worker script can be imported using [new Worker()](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker) and [new SharedWorker()](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker/SharedWorker). Compared to the worker suffixes, this syntax leans closer to the standards and is the **recommended** way to create workers.

PR #126, opened 2025-09-23
Scope edit frame

mattmacf98:
# Description
This is a prototype PR for the scoping work as part of https://viam.atlassian.net/browse/APP-9156. It enables the following features from the scope doc https://docs.google.com/document/d/1ynTWvFPwRYxAhTy9A3Fr7VcW2BmXMmmVbwYMpFZYNww/edit?tab=t.0#heading=h.tcicyojyqi6c

1. shows metadata about frame (position, orientation, geometry) and allows for editing these values via a form
Screenshot 2025-09-23 at 3 16 29 PM

2. allows re-parenting through drag and drop

https://github.com/user-attachments/assets/fca45ca4-3215-49da-b89d-da3508ddfa21

3. allow users to add and delete frames from components

https://github.com/user-attachments/assets/e2691b3a-23d0-4e64-b151-a9a3174710d8

# TODO
There are a couple of todos to take care of to take this from prototype to fully functional
- whenever the frames refresh, the selected 3d object gets completely replaced and this deselects it, we should enable a way to persist selected 3d objects even if they get replaced in the refresh
- right now the geo select just defaults to some hardcoded values, we should allow people to change the actual shape dimensions as well
- I did not implement proper orientation conversion or supporting the different modes we allow people to set orientation in
- I was having issues getting the useAppQuery to work to get the robotpart by id to grab the config, right now I am explicitly calling using the viam client
- all these updates change the config and then we wait for the viewer to refresh and update, we should either figure out a way to force refresh when the configs change or persist a temporary state to model what the eventual network state will be (I did this for a couple ones like position but others like parenting are trickier)
- drag and drop is quite temperamental and I have seen cases where accidental drags can cause weird things in the heirarchy (like a child being the parent of its own parent). We should robustify and maybe switch to using an established dragging library like https://github.com/isaacHagoel/svelte-dnd-action OR switch the UI to be a select dropdown to re-parent nodes.
- there is a bug currently that breaks my tree traversal when there is a frame that references a frame that does not exist or if there is a cyclical parenting loop (this one should be a quick fix)
- there are also a couple other ux tweaks around the input fields and buttons we should change
- a decent amount of refactoring (I jammed everything into useFrames for the prototype, would be worth moving some stuff out of there)
- this protoype does not go into how to update fragments using overrides yet

### DIFF
--- a/src/lib/components/App.svelte
+++ b/src/lib/components/App.svelte
@@ -22,6 +22,8 @@
 		children?: Snippet
 		getLocalPartConfig?: () => unknown
 		setLocalPartConfig?: (config: Struct, partName: string) => void
+		getPartName?: () => string | undefined
+		isDirty?: () => boolean
 	}
 
 	let {
@@ -30,6 +32,8 @@
 		children: appChildren,
 		getLocalPartConfig,
 		setLocalPartConfig,
+		getPartName,
+		isDirty,
 	}: Props = $props()
 
 	const appClient = useViamClient()
@@ -55,13 +59,13 @@
 		getPartName()
 	})
 
-	if (getLocalPartConfig && setLocalPartConfig) {
+	if (getLocalPartConfig && setLocalPartConfig && getPartName && isDirty) {
 		providePartConfig({
 			appEmbeddedPartConfigProps: {
-				isDirty: () => false, //MATTHEW: todo get this from outer context
+				isDirty: () => isDirty(),
 				getLocalPartConfig,
 				setLocalPartConfig,
-				partName: () => partName,
+				partName: () => getPartName(),
 			},
 		})
 	} else {


micheal-parks:
Nice!

Some thoughts:

> whenever the frames refresh, the selected 3d object gets completely replaced and this deselects it, we should enable a way to persist selected 3d objects even if they get replaced in the refresh

I think that we need an edit / monitor mode. Being able to edit while monitoring is going to cause problems. The biggest one that I'm seeing aside from what you describe is that we're also fetching live poses from the motion service. (in that `<Pose>` component). I think that if you're editing frames the robot should be in its null "T-pose" and not moving around or in the current pose that the robot is in.

> I did not implement proper orientation conversion or supporting the different modes we allow people to set orientation in
I was having issues getting the useAppQuery to work to get the robotpart by id to grab the config, right now I am explicitly calling using the viam client

This is another thing that confuses users, especially allowing people to edit quaternions. For now I'd just support orientation vector degrees.

> drag and drop is quite temperamental and I have seen cases where accidental drags can cause weird things in the heirarchy (like a child being the parent of its own parent). We should robustify and maybe switch to using an established dragging library like https://github.com/isaacHagoel/svelte-dnd-action OR switch the UI to be a select dropdown to re-parent nodes.

I'd vote for just making a select dropdown for reparenting for now. DnD is a time sink.

> there are also a couple other ux tweaks around the input fields and buttons we should change

I'd also vote for making the geometry type selector a dropdown since the list will grow.

mattmacf98:
Screenshot 2025-10-01 at 2 30 44 PM
Screenshot 2025-10-01 at 2 31 01 PM

In this latest update, I switched over to using implicit edit vs monitor mode. When a user starts to edit, they will be marked dirty and we will stop fetching updates to poses until they discard or save changes

I also updated the UI to only allow editing world objects which actually have frames (so the fake gripper geo that gets returned from GetGeometries is shown but cannot be changed)

PR #128, opened 2025-09-24
APP 9587 - Allow drag and drop PCD files

micheal-parks:
Allows users to drop in PCD files. 

Also fixes a bug where out of order worker responses would result in PCD files not showing up.

pcd

### DIFF
--- a/src/lib/components/FileDrop.svelte
+++ b/src/lib/components/FileDrop.svelte
@@ -43,6 +43,17 @@
 		const { files } = event.dataTransfer
 
 		for (const file of files) {
+			const ext = file.name.split('.').at(-1)
+
+			if (ext !== '.pcd') {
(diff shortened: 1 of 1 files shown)


src/lib/components/FileDrop.svelte
```diff
@@ -0,0 +1,107 @@
...
+			})
+
+			reader.addEventListener('load', async (event) => {
+				const arrayBuffer = event.target?.result
+
+				if (!arrayBuffer || typeof arrayBuffer === 'string') {
+					return
+				}
+
+				const result = await parsePcdInWorker(new Uint8Array(arrayBuffer))
```
mattmacf98, line 67: since this file drop only does PCD maybe we could rename it to PCDFIleDrop?

PR #135, opened 2025-09-26
APP-9549 - Fix origin frames and geometry offsets

micheal-parks:


src/lib/components/WorldObjects.svelte
```diff
@@ -22,33 +22,23 @@
 </script>
 
 {#each frames.current as object (object.uuid)}
-	<Pose name={object.name}>
+	<Pose
+		name={object.name}
+		parent={object.referenceFrame}
+	>
```
micheal-parks, line 28: Request poses with respect to the parent reference frame now instead of world, helps simplify stuff, especially for @mattmacf98's project

DTCurrie:
Does this also include rendering an axes helper/"coordinate system" at the non-origin frame?

mattmacf98 (approved):

DTCurrie (approved):

PR #145, opened 2025-10-07
Put geometries back in the same world object

micheal-parks:
### Overview

Separating geometries into another world object was a quick fix for not accounting for the "center" field. This is a better fix that reintegrates them into the frame that they belong to.

### DIFF
--- a/src/lib/components/Geometry.svelte
+++ b/src/lib/components/Geometry.svelte
@@ -83,28 +83,28 @@
 			{uuid}
 			bvh={{ enabled: false }}
 		>
-			{#if geometry?.geometryType.case === 'mesh'}
+			{#if geometry.geometryType.case === 'mesh'}
 				{@const mesh = geometry.geometryType.value.mesh as Uint8Array<ArrayBuffer>}
 				{@const meshGeometry = plyLoader.parse(typeof mesh === 'string' ? atob(mesh) : mesh.buffer)}
 				<T
 					is={meshGeometry}
 					{oncreate}
 				/>
-			{:else if geometry?.geometryType.case === 'line' && metadata.points}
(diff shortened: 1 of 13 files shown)


src/lib/components/Geometry.svelte
```diff
@@ -33,95 +33,111 @@
...
 	let geo = $state.raw<BufferGeometry>()
 
 	const oncreate = (ref: BufferGeometry) => {
 		geo = ref
 	}
-
-	const color = $derived(overrideColor ?? metadata.color ?? colors.default)
 </script>
 
 <T
```
micheal-parks, line 70: We introduce a group that represents the frame's offset, the child mesh now accepts the `center` offset

src/lib/components/Geometry.svelte
```diff
@@ -33,95 +33,111 @@
...
-		/>
-	{:else if geometry?.case === 'sphere'}
-		{@const radiusMm = geometry.value.radiusMm ?? 0}
-		<T.SphereGeometry
-			args={[radiusMm * 0.001]}
-			{oncreate}
-		/>
-	{:else if geometry?.case === 'capsule'}
-		{@const { lengthMm, radiusMm } = geometry.value}
+	<AxesHelper
```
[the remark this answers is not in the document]
micheal-parks: Yep!

src/lib/hooks/useFrames.svelte.ts
```diff
@@ -49,25 +49,15 @@ export const provideFrames = (partID: () => string) => {
 				new WorldObject(
 					frameName,
```
mattmacf98, line 50: Thank you!🚀

mattmacf98 (approved):

PR #449, opened 2026-03-03
convert inline worker generation to a post-processing step

DTCurrie:
Web workers using `new Worker(new URL('./worker.js', import.meta.url))` break when consumed as a library because Vite's dep optimizer moves the package to .vite/deps/ [but the worker file doesn't follow](https://github.com/vitejs/vite/issues/21422).

This PR switches from pre-generating inlined worker files in `src/lib` to post-processing `dist/` after `svelte-package`.

- `src/lib/loaders/pcd/index.ts` reverted to standard `new Worker(new URL(...))` syntax
- `scripts/post-process-workers.js` runs after `svelte-package`, bundles workers with `esbuild`, and rewrites `dist/` to use `Blob` URLs
- check/prepare scripts no longer need a pre-build step
- Added `scripts/README.md` documenting both build scripts

The previous approach (pre-generating `worker.inline.ts` before `svelte-package`) required running a build step before check, didn't support HMR for worker changes in dev, and left generated files in the source tree. Post-processing `dist/ `solves all three: dev mode works natively with Vite, CI runs `check` without a pre-build, and the source stays clean.

I tested running locally, and the changes were reflected in HMR. By running `pnpm preview` and `make up` in the environments, I was able to file-drop a PCD and load it.

- [ ] Test downstream with linking


mattmacf98 (approved):

DTCurrie:
@claude rebase this PR and resolve the merge conflicts

PR #533, opened 2026-04-06
add point cloud chunking

DTCurrie:
Adds the server-side chunked-entity store and the client-side chunk pull loop so large point clouds and point sets can be delivered progressively instead of as a single payload through the `DrawService`. Builds on the `Chunks` metadata, `ChunkProgress` trait, and `writeBufferGeometryRange` utility introduced in #544 

### Protos

- New `GetEntityChunk` RPC in `service.proto`, with `GetEntityChunkRequest` (`uuid` + element `start`) and `GetEntityChunkResponse` (`Transform`/`Drawing` oneof, `start`, `done`) messages.

### Go Draw API

- New `chunker.go` defines a generic `DrawableChunker` interface, a `Chunk` struct, and a `ChunkSender` that iterates a chunker and dispatches `AddEntity` for chunk 0 and `UpdateEntity` for the rest, with optional progress callbacks.
- `PointCloudChunker` iterates a `pointcloud.PointCloud` and `PointsChunker` slices `[]r3.Vector`.
- New `disk_buffer.go` provides an append/read file buffer used to back chunk storage.
- `NewDrawService` now takes a `tempDir` and cleans stale files from it on boot; introduces a `chunkedEntity` type and a `chunked` map on `DrawService`.
- `AddEntity` detects `Chunks` metadata and creates a chunked entity with disk-backed buffers for data, colors, and opacities.
- `UpdateEntity` appends subsequent chunks to the entity via `accumulateChunk`.
- New `GetEntityChunk` handler reads a slice from disk and blocks on a `sync.Cond` until enough data is available or the producer signals complete, returning a `done` flag on the last chunk.
- `RemoveEntity`, `RemoveAll`, and `RemoveAllDrawings` close and delete any chunked entity buffers.
- `StreamEntityChanges` replays the current world state to new subscribers; for chunked entities it replays chunk 0 via `buildChunkedReplayMsg` so late joiners get a coherent view.

### Go Client API

- `DrawPointCloudOptions` and `DrawPointsOptions` gain `ChunkSize` and `OnProgress` fields. When `ChunkSize > 0`, `DrawPointCloud`/`DrawPoints` take the chunked path via new `chunkPointCloud`/`chunkPoints` helpers.
- `DrawPointCloud`/`DrawPoints` refactored to share `buildPointCloud`/`buildPoints` helpers across the normal and chunked paths.
- New `server.DrainChunks` drains a `ChunkSender` until `io.EOF` and returns the server-assigned UUID.
- New `resolveTmpDir` locates a `.tmp` directory next to `go.mod` and passes it to `NewDrawService`.

### Frontend

- `useDrawService.svelte.ts` detects chunked drawings on `ADDED` events (`drawing.metadata?.chunks`), registers the `ChunkProgress` trait, and starts a `pullChunks` loop that calls `client.getEntityChunk`, writes each returned chunk into the pre-allocated buffer via `writeBufferGeometryRange`, and updates `ChunkProgress` until `done`. The loop yields to `requestAnimationFrame` between chunks and aborts cleanly on disconnect.
- Small unrelated cleanup in `ArrowGroups.svelte` to dispose of arrow groups before removing them from the entity map.

### Why?

**Why not use actual streaming for this?**

We want the frontend to acknowledge that it has received and finished processing a chunk of data before receiving the next one. Looking into bi-directional streaming, it does not seem super well supported (at least not with the RPC library we are using now), and is more difficult to manage. Instead, we use a chunk/pagination model so the frontend has more control over the pacing.

**Why use a disk buffer?**

I tested this with an actual PCD (~700 MB) that came from a production machine. We persist state in the draw service, meaning we store entity data in memory. For large buffer data, this could quickly eat up a lot of system memory, especially if we wanted to load the PCD with the client _then_ send it to the draw service. So instead, we just write the buffers to disk. We also clean up any stale files we have written each run, to avoid using up a bunch of disk space over time.

### Testing

Ran the `draw-client` e2e suite.

Added two new tests: `draw point cloud in chunks` and `draw points in chunks`.

Screenshot 2026-04-07 at 10 35 32 AM

### PR Stack

1. #544
2. #551 
3. 🎯 #533


client/api/draw_point_cloud.go
```diff
@@ -81,3 +114,20 @@ func DrawPointCloud(options DrawPointCloudOptions) ([]byte, error) {
...
+	pointCloud, drawOptions, err := buildPointCloud(options)
+	if err != nil {
+		return nil, err
+	}
+
+	chunker := draw.NewPointCloudChunker(pointCloud, options.Name, options.ChunkSize, drawOptions...)
+	metadata := &drawv1.Chunks{
+		ChunkSize: uint32(options.ChunkSize),
+		Total:     uint32(pointCloud.PointCloud.Size()),
+		Stride:    3 * 4, // float32 xyz = 12 bytes per point
```
mattmacf98, line 130: cant pcds have colors?

draw/chunker.go
```diff
@@ -0,0 +1,178 @@
...
+		return io.EOF
+	}
+
+	chunk, ok := <-s.streamed
+	if !ok {
+		s.done = true
+		return io.EOF
+	}
+
+	if chunk.IsFirst {
```
mattmacf98, line 105: nit: instead of adding a bool for isFirst can we not just check that start == 0?

draw/draw_service.go
```diff
@@ -33,33 +36,108 @@ type storedEntity struct {
 
 const entitySubscriberBufferSize = 64
 
-// DrawService  stores transforms and drawings keyed by UUID and fans out change events to streaming subscribers.
+type chunkedEntity struct {
+	mu            sync.Mutex
+	cond          *sync.Cond
+	metadata      *drawv1.Chunks
+	data          *diskBuffer
```
mattmacf98: just for my understanding, why are we saving these to disk and not just holding in memory?

Makefile
```diff
@@ -122,7 +122,7 @@ proto-format-check:
 	@pnpm exec buf format --diff --exit-code
 
 .PHONY: proto
-proto: proto-clean proto-vendor 
+proto: proto-vendor
```
mattmacf98, line 125: why remove the clean step?

src/lib/hooks/useDrawService.svelte.ts
```diff
@@ -114,11 +117,111 @@ export function provideDrawService() {
...
+				writeBufferGeometryRange(buffer, positions, response.start, metadata)
+
+				const chunkElements = positions.length / 3
+				nextStart = response.start + chunkElements
+				entity.set(traits.ChunkProgress, { loaded: nextStart, total: totalElements })
+				invalidate()
+
+				if (response.done) break
+
+				await waitForFrame()
```
mattmacf98: I think this waitForFrame is left from testing

Makefile
DTCurrie: it was constantly regening the buf.lock even when no changes were made

src/lib/hooks/useDrawService.svelte.ts
DTCurrie: ah yeah, I copied a lot of the chunking PR code from here

client/api/draw_point_cloud.go
DTCurrie: Yes, we're saying "chunk size is the number of positions worth of data I want."

I added a couple of e2e tests to cover chunking with color data though for safety:

DRAW-POINT-CLOUD-IN-CHUNKS-WITH-PALETTE-darwin
DRAW-POINT-CLOUD-IN-CHUNKS-WITH-PER-POINT-COLORS-darwin

draw/chunker.go
DTCurrie: it was mostly a helper since we reference "is this the first" chunk a lot and it saved on having to do the comparison every time, but admittedly it is not necessary.

draw/draw_service.go
DTCurrie: > Why use a disk buffer?
> 
> I tested this with an actual PCD (~700 MB) that came from a production machine. We persist state in the draw service, meaning we store entity data in memory. For large buffer data, this could quickly eat up a lot of system memory, especially if we wanted to load the PCD with the client then send it to the draw service. So instead, we just write the buffers to disk. We also clean up any stale files we have written each run, to avoid using up a bunch of disk space over time.

Basically trying to avoid dropping gbs worth of data into memory.

DTCurrie:
@claude review these changes

client/api/draw_point_cloud.go
```diff
@@ -33,12 +34,19 @@ type DrawPointCloudOptions struct {
 	// If not provided, the point cloud's color data will be used.
 	Colors []draw.Color
 
+	// ChunkSize controls streaming.
+	// - When > 0, the point cloud is streamed in chunks of this size.
+	// - Otherwise, the point cloud is sent in a single call.
```
DTCurrie: Update this to clarify chunk size === amount of points (not literal memory size of chunk)

micheal-parks (approved):

DTCurrie:
@claude review this PR

micheal-parks (approved):

PR #534, opened 2026-04-06
refactor select tool to provide context

mattmacf98:
# Description

These are some selection changes I wanted to make to enable the heaviness demo attached.

*Main Changes*
1. switched from callback based approach to provide and use context
2. exported floating panel component
3. export createBinaryPCD so the consumer can create PCDs (which we used to do within the tool)
4. add an explicit clear selection button since we cannot remove entities until the consumer is finished with them
5. add a register onEntity Select callback list to allow consumers to inject logic to be executed on entities when they get selected

# Testing

- used the linked version of this motion tools and achieved the UX shown in the video and in theis draft PR https://github.com/viamrobotics/sanding-webapp/pull/56

https://github.com/user-attachments/assets/47901b2f-dc23-4412-b2a8-df51eedfa719

- updated and testes the `/select` demo page

https://github.com/user-attachments/assets/5718f3d1-7c66-453a-b9cc-010b1f4ded47


src/lib/components/Selection/Lasso.svelte
```diff
@@ -48,7 +50,7 @@
 			traits.ScreenSpace,
 			traits.RenderOrder(999),
 			traits.Material({ depthTest: false }),
-			traits.Color({ r: 1, g: 0, b: 0 }),
+			traits.Colors(new Uint8Array([255, 0, 0, 255])),
```
mattmacf98: these color -> colors changes are needed for @DTCurrie 's new colors changes which make points only use colors trait https://github.com/viam-labs/motion-tools/pull/477/changes

src/lib/components/Selection/Ellipse.svelte
```diff
@@ -211,17 +213,21 @@
...
 			traits.RenderOrder(999),
 			traits.Material({ depthTest: false }),
 			traits.Points,
 			traits.Removable,
 			selectionTraits.SelectionEnclosedPoints,
 			selectionTraits.PointsCapturedBy(ellipse)
 		)
+
+		for (const callback of selection.selectionCallbacks) {
+			callback(result)
```
mattmacf98: needed this register and apply callback approach because I needed a way for the consumer to hook into the entity on creation to apply business logic e.x. change the point colors

src/lib/components/Selection/Tool.svelte
```diff
@@ -122,22 +90,5 @@
 {#if isSelectionMode && rect.height > 0 && rect.width > 0}
 	<Ellipse active={selectionType === 'ellipse'} />
 	<Lasso active={selectionType === 'lasso'} />
-
-	<Portal id="dom">
```
mattmacf98, line 126: I have completely moved the management of the UX and functionality for deciding what to do with selected entities to the client

this means the client has slightly more code to right (for the ui panel and to do the post process of the entities themselves) but makes this selection tool much more generic since the action and lifecycle of the entities can be controlled by the consumer now

src/routes/lib/components/selectDemoPage/SelectionOverlay.svelte
```diff
@@ -0,0 +1,48 @@
...
+
+	const rect = new ElementRect(() => dom)
+
+	selectionCtx.registerSelectionCallback((entity: Entity) => {
+		// entity.set(traits.Color, { r: 0, g: 1, b: 0 })
+		entity.set(traits.Colors, new Uint8Array([0, 255, 0]))
+	})
+</script>
+
+<Portal id="dom">
```
mattmacf98, line 24: an example of the extra code consumers have to write, they are responsible for entity cleanup, oncommit entity action and the UI to invoke these functions on the context

src/lib/components/Selection/Lasso.svelte
DTCurrie: just a heads up I think [points](https://github.com/viam-labs/motion-tools/blob/main/src/lib/components/Entities/Points.svelte) reads from `color` and `colors` now just depends on if you want to use a single color or multiple colors.

mattmacf98: nice! just changed back to using color trait now

src/lib/components/Selection/Ellipse.svelte
```diff
@@ -24,6 +25,7 @@
 	const world = useWorld()
 	const controls = useCameraControls()
 	const { scene, dom, camera } = useThrelte()
+	const selection = useSelection()
```
micheal-parks: I'd maybe call this something else to avoid complete overlap with the main `useSelection` hook, especially when doing search

micheal-parks: `useSelectionTool`? `useSelectionPlugin`?

src/lib/components/Selection/useSelection.svelte.ts
```diff
@@ -0,0 +1,46 @@
...
+			return callbacks
+		},
+		clearSelections() {
+			for (const entity of world.query(selectionTraits.SelectionEnclosedPoints)) {
+				if (world.has(entity)) {
+					entity.destroy()
+				}
+			}
+		},
+		registerSelectionCallback(callback) {
```
micheal-parks: This feels like circumventing reactivity, and potentially causing too much complexity. Why doesn't this hook keep a reactive array of entities in `$state`, which then any consumer would read from? ex:

```ts
const selectionContext = useSelection()

$effect(() => {
  const entity = selectionContext.entities.at(-1)
  entity.set(traits.Color, { r: 0, g: 1, b: 0 })
})

File: src/lib/components/Selection/Ellipse.svelte (hunk 5 above)
Comment (mattmacf98) [reply src/lib/components/Selection/Ellipse.svelte]: I'll update to useSelectionPlugin

File: src/lib/components/Selection/useSelection.svelte.ts (hunk 6 above)
Comment (mattmacf98) [reply src/lib/components/Selection/useSelection.svelte.ts]: ok yeah I can update to have the consumer own the reactivity logic

File: src/lib/components/Selection/Ellipse.svelte [hunk 7]
```diff
@@ -225,9 +225,7 @@
 			selectionTraits.PointsCapturedBy(ellipse)
 		)
 
-		for (const callback of selection.selectionCallbacks) {
-			callback(result)
-		}
+		selection.addEntity(result)
```
micheal-parks: I'm wondering if we even need an addEntity function, could the `useSelectionPlugin` not have this?

```ts
const entities = useQuery(trait.sSelectionEnclosedPoints)
```

I think this would be a bit safer due to not needing to manually manage the array

micheal-parks (approved):

src/lib/components/Selection/Ellipse.svelte
mattmacf98: my only issue with that is it will trigger on removal as well, whereas the explicit one would not update on removing

but I can just account for that in the consumer side and keep this simple to use the query instead

PR #536, opened 2026-04-07
add metadata to show axes helper

DTCurrie:
Adds a new `show_axes_helper` metadata field that lets Go clients control whether the axes indicator is rendered on each entity. Previously, axes helpers were unconditionally shown for transforms and hidden for drawings, with no way to override from the backend.

### Proto

- Added `optional bool show_axes_helper = 4` to `Metadata` — controls whether the axes indicator is rendered on the entity. Defaults to `false`.

### Go Draw API

- `DrawConfig` gains a `ShowAxesHelper` field, defaulting to `true` in `NewDrawConfig`.
- `WithAxesHelper(bool)` is a new `DrawableOption` that overrides the default.
- `WithMetadataAxesHelper(bool)` is a new `DrawMetadataOption`.
- `Metadata` gains a `ShowAxesHelper` field; `NewMetadata`, `Metadata.ToProto`, `MetadataToStruct`, and `StructToMetadata` all propagate it.

### Go Client API

- `Attrs` gains `ShowAxesHelper *bool`. When non-nil, passes `draw.WithAxesHelper(...)` to the draw layer, overriding the `DrawConfig` default.

### Frontend

- `metadataFromStruct` in `metadata.ts` now extracts `show_axes_helper` from struct fields.
- `isMetadataField` recognizes `show_axes_helper`.
- `drawTransform` / `updateTransform` read `showAxesHelper` from parsed metadata instead of from an `options` parameter. The `showAxesHelper` option is removed from the `Options` type.
- `drawDrawing` / `updateDrawing` read `showAxesHelper` from the drawing's typed metadata.

### Why?

**Why default to `true` in `draw/` and let `client/api/` override via `Attrs`?**

The `draw/` package defaults `ShowAxesHelper` to `true` in `NewDrawConfig` because most entities benefit from visible axes indicators. The `client/api/` layer can override this per-entity by setting `Attrs.ShowAxesHelper` to `false`. Since the field lives in `Attrs`, adding it required zero changes to the seven `Draw*Options` structs — it was a two-line addition in `attrs.go`.

### Testing

Ran the `draw-client`, `go-client`, and `snapshot` e2e tests.


protos/draw/v1/metadata.proto
```diff
@@ -9,4 +9,11 @@ message Metadata {
   // For simple objects, this is a single color of [r, g, b, a]
   // For complex objects, this is an array of colors of [r, g, b, a, ...]
   optional bytes colors = 1;
+
+  reserved 2;
+  reserved 3;
```
DTCurrie: reserving these for #519

client/api/api.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+var (
+	// DefaultTransformShowAxesHelper is the default value for the ShowAxesHelper option for transforms.
+	DefaultTransformShowAxesHelper = true
+
+	// DefaultDrawingShowAxesHelper is the default value for the ShowAxesHelper option for drawings.
+	DefaultDrawingShowAxesHelper = false
```
mattmacf98: do we really want to have the default be no axis for drawings still? it would be nice if there is less diff between transform and drawings

client/api/draw_geometries.go
```diff
@@ -28,6 +28,10 @@ type DrawGeometriesInFrameOptions struct {
 	// distance to one another are culled, reducing the total point count and improving
 	// rendering performance. Set to 0 (default) to disable downscaling.
 	DownscalingThreshold float64
+
+	// ShowAxesHelper controls whether the axes helper (RGB XYZ indicator) is shown on each entity.
+	// If nil, defaults to DefaultTransformShowAxesHelper.
+	ShowAxesHelper *bool
```
mattmacf98: its a shame adding generic things like this require so many changes,
wondering if we could have these options inherit from a common options thing with things like this in it and those have a generic `applyGenericOptions` function, then we would only have to change there and could leave the actual geometries alone when we add things like this

mattmacf98 (approved):

client/api/api.go
DTCurrie: This is how we have it currently. The question is whether we should render axes helpers for drawings. We could just simplify this to always default to true, and you have to specify false to hide it. @micheal-parks thoughts?

client/api/draw_geometries.go
DTCurrie: I actually had something like this with a `BuildMetadata` thing on the drawable object, but it was still kind of annoying to add everywhere. I will take another pass to see if there is a way to simplify this. might put up something as a pre-cursor to this PR to just cleanup general metadata handling and export some of the utils we talked about for getting color info from them.

DTCurrie: ok added this #541 

will apply that to this and the invisible PR and see how much it reduces the diff.

PR #541, opened 2026-04-08
rework metadata

DTCurrie:
Reworks the metadata plumbing across the Go Draw API and the frontend to reduce boilerplate at every call site, and standardizes the Go client API option structs so all single-object draw functions share a consistent set of entity fields.

### Go Draw API

- `DrawConfig.BuildMetadata(opts ...DrawMetadataOption)` centralizes metadata construction.
- `DrawMetadataOption` is now exported.
- `NewDrawing` accepts `(*DrawConfig, Shape, ...DrawMetadataOption)` instead of individual UUID/name/parent/pose/metadata arguments.
- `NewTransform` accepts `(*DrawConfig, Geometry, ...DrawMetadataOption)` instead of individual arguments.
- `MetadataToStruct` no longer returns an error (it never could fail).
- All five shape `Draw()` methods (arrow, line, points, nurbs, model) now call `NewDrawing(config, shape, ...)` instead of unpacking config fields.
- `DrawnGeometry.Draw` and `DrawnPointCloud.Draw` use the simplified `NewTransform(config, geometry, ...)` signature.
- `Snapshot.DrawFrame` accepts `...DrawMetadataOption` instead of a raw `*structpb.Struct`.

### Go Client API

- New `Attrs` struct in `attrs.go` holds common attributes for all `Draw*` calls. Future metadata fields (e.g. `ShowAxesHelper`, `Invisible`) are added here once instead of to every option struct.
- New `entityAttributes(id, parent, attrs)` helper centralizes the conversion from client API fields to `draw.DrawableOption`, replacing duplicated inline logic across all seven single-object draw functions.
- Each `Draw*Options` struct gains an `Attrs *Attrs` field for optional attribute overrides.
- `DrawGeometryOptions` gains a `Name` field that overrides the geometry label.
- `DrawPointCloudOptions` replaces `Label` with `Name` and adds a `Parent` field, matching the other option structs.
- All seven `Draw*` functions now delegate to `entityAttributes()` instead of building `DrawableOption` slices inline.

### Frontend

- `Metadata` type is now `PlainMessage<MetadataProto>`, aligning with the proto definition.
- `parseMetadata` renamed to `metadataFromStruct` and exported from `lib.ts`.
- `createBufferGeometry` and `updateBufferGeometry` now derive the color attribute stride from `metadata.colorFormat` via `colorStride`.
- Removed `Uint8Array<ArrayBuffer>` / `Float32Array<ArrayBuffer>` generics throughout in favour of bare `Uint8Array` / `Float32Array` that he API uses.

### Testing

Ran the `draw-client`, `go-client`, `snapshot`, and `file-drop` e2e suites.


DTCurrie:
@mattmacf98 added this to reduce the boilerplate needed to add #536 and #537 

Would like to merge all three before I add the chunking logic for PCDs from the world state store service for the hackathon

mattmacf98 (approved):

PR #544, opened 2026-04-10
add chunking

DTCurrie:
Adds progressive/chunked delivery of large point clouds and point sets so they can be streamed incrementally instead of sent as a single payload. The frontend pre-allocates geometry from `Chunks` metadata, then pulls remaining chunks over `get_entity_chunk` DoCommands on the world-state store, updating a visible progress indicator as they arrive.

### Protos

* `metadata.proto` - new `Chunks` message with `chunk_size`, `total`, and `stride`, plus an optional `chunks` field on `Metadata`. Regenerated `draw/v1/metadata.pb.go`.

### Frontend

**Chunking core**

* `src/lib/chunking.ts` - new `createChunkLoader` that takes a `fetchChunk` callback, pulls chunks one-at-a-time, writes positions/colors/opacities into the entity's existing `BufferGeometry`, and updates/removes the `ChunkProgress` trait.
* `src/lib/attribute.ts` - adds `preAllocateBufferGeometry` and `writeBufferGeometryRange`.
* `src/lib/ecs/traits.ts` - new `ChunkProgress` trait (`{ loaded, total }`).
* `src/lib/metadata.ts` - parses `chunks` out of the `Struct` metadata into `{ chunkSize, total, stride }`.

**Draw / world-state integration**

* `src/lib/hooks/useWorldState.svelte.ts` - wires up `createChunkLoader` with a `get_entity_chunk` DoCommand adapter (decodes the base64 positions/colors/opacities from the JSON `Struct` response), calls `chunkLoader.start` when an entity spawns, and disposes it on teardown. 

**Rendering / UI**

* `src/lib/components/overlay/left-pane/TreeNode.svelte` — new `progressIndicator` snippet that renders a conic-gradient progress ring next to the visibility toggle whenever the entity has `ChunkProgress`.

### Why?

**Why chunk through `get_entity_chunk` instead of true streaming?**

We want the frontend to acknowledge each chunk (write it to the buffer, yield a frame, update progress) before requesting the next one. Pulling over DoCommand keeps the round-trip handshake explicit and avoids saturating the main thread on multi-hundred-MB point clouds.

### Testing

Ran all the e2e tests including new world state store e2e tests from #551 

Screenshot 2026-04-20 at 12 27 26 PM

### PR Stack

1. 🎯 #544
2. #551 
3. #533


src/lib/components/Entities/Mesh.svelte
DTCurrie, line 1: Changes here are to support updating opacities, which wasn't working. Not sure if this is the best way to handle this.

src/lib/components/Entities/Pose.svelte
DTCurrie, line 1: Returns the edited pose while we wait for config updates to be resolved on the machine.

src/lib/hooks/useFrames.svelte.ts
DTCurrie, line 1: This adds some statefulness around when we have a pending save from editing a config. I noticed in our e2e tests, you could save a frame, then refresh the page or something, and you would end up getting stale data from the machine. This made our e2e tests flaky, and also led to some unintended rendering issues where things did not render where expected while we were in a "waiting for the machine to resolve frame updates" state.

src/lib/hooks/useFrames.svelte.ts
```diff
@@ -122,22 +180,19 @@ export const provideFrames = (partID: () => string) => {
...
 				const pose = createPose(frame.poseInObserverFrame?.pose)
+
 				const center = frame.physicalObject?.center
 					? createPose(frame.physicalObject.center)
 					: undefined
 				const resourceName = currentResourcesByName[frame.referenceFrame]
-				const color = resourceNameToColor(resourceName)
+				const color =
+					resourceNameToColor(resourceName) ??
+					subtypeToColor(currentComponentSubtypeByName[frame.referenceFrame])
```
DTCurrie, line 189: better default color handling, I was seeing colors flash in e2e tests

micheal-parks:
@DTCurrie would it be possible to get the "Pending-save coordination" stuff into a separate PR to merge faster since I started work on frames?

DTCurrie:
> @DTCurrie would it be possible to get the "Pending-save coordination" stuff into a separate PR to merge faster since I started work on frames?

sure

src/lib/components/Entities/Arrows/Arrows.svelte
```diff
@@ -37,7 +37,7 @@
 	})
 </script>
 
-<Portal id={parent.current}>
+<Portal id={parent.current ?? 'world'}>
```
mattmacf98: don't we already account for null parents with this portal https://github.com/viam-labs/motion-tools/blob/main/src/lib/components/Scene.svelte#L103?

src/lib/components/Entities/Mesh.svelte
```diff
@@ -56,8 +56,21 @@
...
 
-	const mesh = new Mesh()
+	let material = $state.raw<Material>(new Material())
+	let lastTransparent = false
+	$effect(() => {
+		const isTransparent = currentOpacity < 1
+		material.transparent = isTransparent
+		material.depthWrite = !isTransparent
+		material.opacity = currentOpacity
+		if (lastTransparent !== isTransparent) {
```
mattmacf98: nit: isn't lastTransparent just materil.transparent so we can do this


```
const isTransparent = currentOpacity < 1
material.depthWrite = !isTransparent
material.opacity = currentOpacity
if (material.transparent !== isTransparent) {
   material.transparent = isTransparent
   material.needsUpdate = true
   invalidate()
}
```

src/lib/chunking.ts
```diff
@@ -0,0 +1,103 @@
...
+				})
+
+				const chunkElements = chunk.positions.length / 3
+				nextStart = chunk.start + chunkElements
+				entity.set(traits.ChunkProgress, { loaded: nextStart, total })
+				invalidate()
+
+				if (chunk.done) break
+
+				await waitForFrame()
```
mattmacf98: isn't this adding artificial load time? why should we wait for the next animation frame before starting the next chunk load?

src/lib/chunking.ts
```diff
@@ -0,0 +1,103 @@
...
+			}
+		}
+	}
+
+	return {
+		start(uuid: string, entity: Entity, metadata: Metadata) {
+			const chunks = metadata.chunks
+			if (!chunks || chunks.total <= 0) return
+
+			entity.add(traits.ChunkProgress({ loaded: chunks.chunkSize, total: chunks.total }))
```
mattmacf98, line 91: why does progress start at chunkSize and not 0?

src/lib/draw.ts
```diff
@@ -133,8 +138,7 @@ export const updateTransform = (
 ): void => {
 	entity.set(traits.Pose, createPose(poseInObserverFrame?.pose))
 
-	const parent = poseInObserverFrame?.referenceFrame
-	if (parent && parent !== 'world') entity.set(traits.Parent, parent)
```
mattmacf98: was there a reason we did not set the trait when it was world before?

src/lib/chunking.ts
DTCurrie: ah yeah this was leftover from some debugging

src/lib/chunking.ts
DTCurrie: Bby the time we know we are dealing with a chunked entity, we have already received the first chunk

src/lib/draw.ts
DTCurrie: I actually will probably revert this. I was seeing some issue updating transforms from the world state store related to the parent trait state, especially when it came to editing frames to change their parent. We originally decided that "lack of parent trait === 'world'" but I am not sure we are being consistent with it which may be causing some issues.

src/lib/components/Entities/Arrows/Arrows.svelte
DTCurrie: see my other comment but may revert this, I think we have some inconsistencies with parent trait management.

src/lib/components/Entities/Mesh.svelte
DTCurrie: Ah yeah I was just storing it in state before deciding to move to a material, will just use the material.

src/lib/attribute.ts
```diff
@@ -56,3 +56,67 @@ export const updateBufferGeometry = (
 		}
 	}
 }
+
+export const preAllocateBufferGeometry = (
```
micheal-parks, line 60: These two functions are very similar to `createBufferGeometry` and `updateBufferGeometry`, it would be nice in the future to merge them, but not blocking here

DTCurrie: Yeah, I didn't want to mess with the existing code, and these are also only built for specific positions where I think we could supply the values and stride so they can be used to write any sort of buffer data.

mattmacf98 (approved):

PR #545, opened 2026-04-13
Feat add interaction layer trait

mattmacf98:
# Description
https://viam.atlassian.net/browse/APP-15947
This PR allows us to specify if an entity is eligible for lassoing, this will allow sanding to add multiple PCDs to their scene without having to worry about selecting points on the non-hardtop pcd

# Testing
- ran locally with the /selection endpoint and saw we can turn on and off select-ability for PCDs 

https://github.com/user-attachments/assets/b606d814-b843-4859-85fd-84609bbb4945

### DIFF
--- a/src/lib/components/PCD.svelte
+++ b/src/lib/components/PCD.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
 	import type { ConfigurableTrait, Entity } from 'koota'
 
+	import type { InteractionLayerValue } from '$lib/ecs/traits'
+
 	import { createBufferGeometry } from '$lib/attribute'
 	import { ColorFormat } from '$lib/buf/draw/v1/metadata_pb'
 	import { traits, useWorld } from '$lib/ecs'
@@ -8,12 +10,15 @@
 
 	interface Props {
 		data: Uint8Array
-		name?: string
-		renderOrder?: number
-		oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
+		options?: {
+			name?: string
+			renderOrder?: number
+			interactionLayers?: InteractionLayerValue[]
+			oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
+		}
 	}
 
-	let { data, name, renderOrder, oncreate }: Props = $props()
+	let { data, options }: Props = $props()
 
 	const world = useWorld()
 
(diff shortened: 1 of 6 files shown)


src/lib/ecs/traits.ts
```diff
@@ -163,6 +163,12 @@ export const DotSize = trait(() => 10)
 
 export const ReferenceFrame = trait(() => true)
 
+/**
+ * Interaction layers for entities
+ */
+export type InteractionLayerValue = 'selectTool'
+export const SelectToolInteractionLayer = trait(() => true)
```
micheal-parks, line 170:
```suggestion
export const InteractionLayer = trait(() => true)
```
This I'd wager will have more general application than the select tool

src/lib/components/PCD.svelte
```diff
@@ -1,19 +1,24 @@
...
+		options?: {
+			name?: string
+			renderOrder?: number
+			interactionLayers?: InteractionLayerValue[]
+			oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
+		}
 	}
 
-	let { data, name, renderOrder, oncreate }: Props = $props()
+	let { data, options }: Props = $props()
```
micheal-parks:
```suggestion
	let { data, ...rest }: Props = $props()
```

src/lib/components/PCD.svelte
```diff
@@ -1,19 +1,24 @@
...
 	interface Props {
 		data: Uint8Array
-		name?: string
-		renderOrder?: number
-		oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
+		options?: {
+			name?: string
+			renderOrder?: number
+			interactionLayers?: InteractionLayerValue[]
+			oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
```
micheal-parks: Since this is a component, I'd instead do a rest param, otherwise you're triggering updates for any one prop change
```suggestion
		name?: string
		renderOrder?: number
		interactionLayers?: InteractionLayerValue[]
		oncreate?: (positions: Float32Array, colors: Uint8Array | undefined) => void
```

src/lib/ecs/traits.ts
mattmacf98: yes but I was thinking of having a bool trait for each like we do for the frame vs draw API ones.

I originally tried to do with a Set<string> in the data and then doing a filter post retrieval, but that is not very efficient in koota

src/lib/components/PCD.svelte
mattmacf98: ok i'll switch that back

src/lib/components/PCD.svelte
mattmacf98: not needed now that I switched back to remove options

src/lib/ecs/traits.ts
micheal-parks: Ah! Ok, I see what you're doing now, misread this

micheal-parks (approved):

PR #551, opened 2026-04-17
Add e2e machine setup

DTCurrie:
Reworks the Playwright e2e setup so each run provisions its own ephemeral Viam machine under a `Viam Viz E2E` org instead of depending on a long‑lived shared machine, and adds a `world-state-store` test module + test suite to exercise the new chunked point cloud path end‑to‑end.

### Frontend

- `e2e/setup.sh` - new bootstrap script. Installs `viam-server`, builds the `world-state-store` Go test module into `e2e/.bin/`, verifies `viam` CLI auth, resolves the `Viam Viz E2E` org, and rotates a fresh org API key into `e2e/.env.e2e`.
- `e2e/helpers/global-setup.ts` - new Playwright `globalSetup`. Connects to `app.viam.com`, finds/creates the `e2e-tests` location, creates a unique `e2e-<username>-<timestamp>` machine, mints a part secret + machine API key, pushes a `bind_address :9090` config, spawns `viam-server` against the generated `viam-e2e.json`, and waits for the part to report recent `lastAccess` before tests start. Returns a `teardown` that stops the server, deletes the machine, and removes the credentials file.
- `e2e/helpers/e2e-config.ts` - new `loadE2EConfig` helper that parses `e2e/.env.e2e` and surfaces a friendly error pointing at `./setup.sh` when it's missing.
- `playwright.config.ts` - wires `globalSetup` to `e2e/helpers/global-setup.ts` and adds `VITE_CONFIGS: '{}'` to the `webServer` env so the app boots without a baked‑in connection config.
- `package.json` - `test:e2e` and `test:e2e-ui` now run `./e2e/setup.sh` before `playwright test` so contributors get a provisioned binary/API key automatically.
- `.gitignore` - ignores `e2e/.bin/` and `e2e/.env.e2e`.

### E2E fixtures

- `e2e/fixtures/with-robot.ts` - new Playwright fixture replacing the old `e2e/fixtures.ts`. Reads env vars set by global setup, writes the per‑machine connection config into IndexedDB via `injectMachineConfig`, reloads, and retries up to 5× waiting for the `Machine connection configs` button to show `live`. Exposes `robotPage` with `page`, `config`, `viamClient`, `takeScreenshot`, and `assertScreenshots`, plus a `connectOrgViamClient` helper for org‑scoped operations (fragments).
[...]

PR #555, opened 2026-04-21
parent trait handling

DTCurrie:
Ports the parent-trait-related subset of #544 so this can be reviewed without the rest of the chunking work. Codifies that a world-parented entity intentionally carries no `Parent` trait and fixes a real regression where editing a frame's parent back to `'world'` left a stale trait behind.

### Frontend

- `src/lib/ecs/traits.ts` - new `getParentTrait(parent)` spawn helper that returns `[]` for `undefined`, `''`, or `'world'` and `[Parent(parent)]` otherwise.
- `src/lib/draw.ts`:
  - `drawTransform`, `drawDrawing`, `drawModel` spawn via `...getParentTrait(poseInObserverFrame?.referenceFrame)`. 
  - `updateTransform` and `updateDrawing` use `traits.setParentTrait(entity, poseInObserverFrame?.referenceFrame)`.
- `src/lib/hooks/useDrawAPI.svelte.ts` - switched to `setParentTrait(existing, parent)`. `drawFrames` and `drawGeometry` use `getParentTrait(...)`.
- `src/lib/hooks/useFrames.svelte.ts` - spawn path uses `getParentTrait(parent)`.
- `src/lib/hooks/usePointcloudObjects.svelte.ts` - uses `getParentTrait(...)`.
- Portal consumers - `Arrows.svelte`, `Frame.svelte`, `GLTF.svelte`, `Geometry.svelte`, `Line.svelte`, `Points.svelte` all change `<Portal id={parent.current}>` to `<Portal id={parent.current ?? 'world'}>`, so entities without a `Parent` trait explicitly route to the `"world"` `PortalTarget`.

### Why?

**Why replace `.set(traits.Parent, parent)` with `setParentTrait`?**

Two bugs in the old pattern. First, `.set` on an entity that never had the trait silently no-ops in koota (no add fallback), so the first-ever parent assignment was lost. Second, and more visibly, the old code never removed `Parent` when the new value was `'world'`, so editing a frame's parent back to world left a stale trait that `buildTree` and the portal consumers then disagreed about. `setParentTrait` centralizes "add / update / remove" in one place.

**Why `parent.current ?? 'world'` in every portal consumer?**

Threlte's `<Portal>` defaults `id` to `'default'` when `id` is `undefined`. Before this PR, world-parented entities were being routed to the bare `<PortalTarget />` (which has `id='default'`), while entities with an explicit `'world'` parent. Consumers like `BatchedArrows` that already did `?? 'world'` went through `<PortalTarget id="world" />`. 

### Testing

Unit tests passed. Ran e2e tests with code merged down to #551

### DIFF
--- a/src/lib/components/Entities/Arrows/Arrows.svelte
+++ b/src/lib/components/Entities/Arrows/Arrows.svelte
@@ -37,7 +37,7 @@
 	})
 </script>
 
-<Portal id={parent.current}>
+<Portal id={parent.current ?? 'world'}>
 	<T
 		is={arrows}
 		name={entity}
--- a/src/lib/components/Entities/Arrows/Arrows.svelte
+++ b/src/lib/components/Entities/Arrows/Arrows.svelte
@@ -37,7 +37,7 @@
 	})
 </script>
 
-<Portal id={parent.current ?? 'world'}>
+<Portal id={parent.current}>
 	<T
 		is={arrows}
 		name={entity}
--- a/src/lib/components/Scene.svelte
+++ b/src/lib/components/Scene.svelte
@@ -100,6 +100,7 @@
 
 	<T.Group attach={focusedObject ? false : undefined}>
 		<PortalTarget id="world" />
+		<!-- Should we keep this? -->
 		<PortalTarget />
 
 		<Entities />
--- a/src/lib/components/Scene.svelte
+++ b/src/lib/components/Scene.svelte
@@ -99,8 +99,6 @@
 	{/if}
 
 	<T.Group attach={focusedObject ? false : undefined}>
-		<PortalTarget id="world" />
-		<!-- Should we keep this? -->
 		<PortalTarget />
 
 		<Entities />
--- a/.changeset/parent-trait-consistency.md
+++ b/.changeset/parent-trait-consistency.md
@@ -1,0 +1,5 @@
+---
+'@viamrobotics/motion-tools': patch
+---
+
+Consolidate `Parent` trait handling.
--- a/src/lib/__tests__/draw.spec.ts
+++ b/src/lib/__tests__/draw.spec.ts
@@ -17,9 +17,10 @@ import {
 } from '$lib/buf/draw/v1/drawing_pb'
 import { Metadata } from '$lib/buf/draw/v1/metadata_pb'
 import { traits } from '$lib/ecs'
+import { getParentTrait, setParentTrait } from '$lib/ecs/traits'
 import { createPose } from '$lib/transform'
 
-import { drawDrawing, drawTransform } from '../draw'
+import { drawDrawing, drawTransform, updateTransform } from '../draw'
 
 describe('drawTransform', () => {
 	let world: World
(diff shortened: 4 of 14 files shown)


src/lib/components/Scene.svelte
```diff
@@ -100,6 +100,7 @@
 
 	<T.Group attach={focusedObject ? false : undefined}>
 		<PortalTarget id="world" />
+		<!-- Should we keep this? -->
 		<PortalTarget />
```
mattmacf98, line 102: if we have gotten all the places we portal from with `?? 'world'` then no need to keep

DTCurrie: yeah unless someone knows of a specific reason to keep it I think it can be removed

src/lib/components/Entities/Arrows/Arrows.svelte
```diff
@@ -37,7 +37,7 @@
 	})
 </script>
 
-<Portal id={parent.current}>
+<Portal id={parent.current ?? 'world'}>
```
micheal-parks: This adds quite a bit more verbosity - why was this changed? There wasn't a bug here AFAIK, the main scene component has these two portals as siblings:

```svelte
		<PortalTarget id="world" />
		<PortalTarget />
```

mattmacf98 (approved):

src/lib/components/Scene.svelte
micheal-parks: I'd argue that we should keep this (and not nullish coalesce into world everywhere) - this portaltarget follows the idea of implicitly having the "no parent means object is in world space" behavior. I'd either keep this and not nullish coalesce "world" values or make parent required. An in-between design would likely cause objects to occasionally not show up due to developer error IMO

src/lib/components/Entities/Arrows/Arrows.svelte
DTCurrie: I was seeing bugs while getting e2e tests working, especially around editing frames and changing parents, and I noticed we were not consistently following the rules for handling the parent trait and the world.

Is there a reason to have the sibling portal targets? I am fine removing the nullish coalescence and saying everything with a parent of `world` goes to the `default` portal target, we just need to make a decision here and be consistent.

src/lib/components/Scene.svelte
DTCurrie: Why not just never allow the `world` parent trait, which we were not doing consistently, and not have the `<PortalTarget id="world" />` at all? We just treat a parent of `world` as "portal to the default portal target".

src/lib/components/Entities/Arrows/Arrows.svelte
micheal-parks: No reason, I think your suggestion to only have the default PortalTarget is the winner

src/lib/components/Scene.svelte
micheal-parks: I like this ^^

micheal-parks (approved):

PR #569, opened 2026-04-23
world state rendering fixes

DTCurrie:
During the build on Viam, I noticed issues with point cloud rendering and updating when using a world state store service. Point clouds streamed from a world state store now render as actual points (not a triangulated mesh with an edges-geometry wireframe drawn on top), and the point cloud entity correctly re-renders its positions and colors when the underlying `physicalObject` or `metadata` changes.

### Frontend

* `src/lib/components/Entities/Entities.svelte` - added `Not(traits.Points, traits.LinePositions, traits.GLTF)` to the `worldStateEntities` query. 
* `src/lib/ecs/traits.ts` - `updateGeometryTrait`'s `pointcloud` case was a silent no-op and is now wired to a new `updatePointCloud` helper.
* `src/lib/draw.ts` - `updateMetadata` for pointclouds now also calls `setColorTraits(entity, colors)` in addition to `updatePointCloudColors`.

### Why?

**Why replace the `BufferGeometry` instead of mutating it when the point count changes?**

Mutating attributes in place via `updateBufferGeometry` leaves the geometry's cached `boundingBox`, `boundingSphere`, and any attached `three-mesh-bvh` `BoundsTree` pointing at stale position data. Frustum culling and accelerated raycasting then read old data, which produced visible rendering artifacts or caused points not to render unexpectedly. Swapping to a fresh `BufferGeometry` lets threlte re-mount the child of the Points object and rebuild downstream bounds/BVH caches cleanly.

**Why does `updateMetadata` now set `traits.Color` even though `updatePointCloudColors` already updates the buffer?**

The new `updatePointCloud` helper in `traits.ts` runs asynchronously (after the PCD worker finishes) and needs to know the authoritative single color to re-expand into the resized color buffer. Keeping `traits.Color` in sync during every metadata update means the async callback can read it back without having to route the metadata through `updateGeometryTrait`.

### Testing

Ran the updated e2e suite.


src/lib/ecs/traits.ts
```diff
@@ -244,5 +247,68 @@ export const updateGeometryTrait = (entity: Entity, geometry?: ViamGeometry) =>
...
+					const fresh = createBufferGeometry(parsed.positions, {
+						colors,
+						colorFormat: ColorFormat.RGB,
+					})
+					buffer.dispose()
+					entity.set(BufferGeometry, fresh)
+				}
+
+				return
+			}
```
DTCurrie, line 298: This is definitely not optimal, but getting buffer geo updates to work will require re-evaluating some things to ensure these render correctly, which made the PR much broader in scope. I think we can focus on optimizing this behavior in a follow-up PR, but I would prefer to get this in now, since the current behavior is broken and I don't think many users are updating point clouds through the world state store right now.

DTCurrie:
merging to do some testing with the build on viam roomba project

PR #596, opened 2026-05-01
useInputMap for user input

micheal-parks:
Replaces our current keyboard input pipeline with the new [`useInputMap` hook](https://threlte.xyz/docs/reference/extras/use-input-map/) from `@threlte/extras`, which is more flexible, powerful, and reduces footprint a bit. Users can also now navigate the scene using a gamepad with practically no extra code, so that's cool!


src/lib/hooks/useEnvironment.svelte.ts
```diff
@@ -5,6 +5,7 @@ export const ENVIRONMENT_CONTEXT_KEY = Symbol('environment')
 interface Environemnt {
 	viewerMode: 'edit' | 'monitor' | 'focus'
 	isStandalone: boolean
+	inputBindingsEnabled: boolean
```
micheal-parks, line 8: This makes more sense to put in the environment hook since it should not be persisted

DTCurrie (approved):

PR #598, opened 2026-05-01
add opacity slider

DTCurrie:
Screenshot 2026-05-01 at 11 19 13 AM

Adds an opacity slider to the details panel for quick tweaking.


micheal-parks (approved):
Since this is a power user feature as a follow up I'd like to stick this - and future stuff like this - in a tab within details

DTCurrie:
> Since this is a power user feature as a follow up I'd like to stick this - and future stuff like this - in a tab within details

Agreed, I actually _almost_ did that but wasn't sure how we wanted to split it up.

PR #640, opened 2026-05-05
Feat add color prop pcd

mattmacf98:
# Description
for this https://viam.atlassian.net/browse/APP-16320 we want to expose a way for users to override PCD colors using props. This pr exposes a way to do that via props

# Testing
- ran SOW app with linked version and made sure I could override the PCD component color with the color prop and when absent showed the real pcd color
Screenshot 2026-05-05 at 2 36 52 PM

### DIFF
--- a/.changeset/eager-bushes-drop.md
+++ b/.changeset/eager-bushes-drop.md
@@ -1,0 +1,5 @@
+---
+'@viamrobotics/motion-tools': minor
+---
+
+add pcd color prop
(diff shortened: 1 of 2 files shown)


DTCurrie:
I am cool with this. A more future-proof solution may be to just accept a metadata object. But that's a much bigger scope than this change.

DTCurrie (approved):

PR #647, opened 2026-05-06
feat(selection): auto-open Details panel when lasso/ellipse selection…

vijayvuyyuru:
Hey yall this change opens the details panel for a selection upon making it. 
I want this change as I am making a tutorial for sanding operators to use the targeting flow. Furthermore, I think this change makes sense as the operator is typically configuring heaviness immediately after making a selection anyway.


src/lib/components/Selection/Tool.svelte
```diff
@@ -43,6 +46,18 @@
 		}
 	})
 
+	// Each time a new entity appears in the selection set, point
+	// `selectedEntity` at it so the Details panel reflects the latest selection.
+	let previousEntities: Entity[] = []
```
vijayvuyyuru, line 51: I want this change as I am making a tutorial for sanding operators to use the targeting flow. Furthermore, I think this change makes sense as the operator is typically configuring heaviness immediately after making a selection anyway.

mattmacf98 (approved):

PR #649, opened 2026-05-06
Use Koota relations for parent / child interactions

micheal-parks:
This PR simply replaces the `Parent()` trait with a `ChildOf()` relation across the entire app. Any optimizations that we discussed as a result of this will come in followups.


src/lib/ecs/useTarget.svelte.ts
```diff
@@ -0,0 +1,50 @@
+import { type Entity, $internal as internal, type Relation, type Trait, type World } from 'koota'
```
micheal-parks, line 1: useTarget is part of the upcoming koota/svelte package. Ported it over since it's not merged into koota yet.

DTCurrie (approved):

PR #661, opened 2026-05-07
Use matrix traits as transform source-of-truth

micheal-parks:
This PR replaces our current pose traits with matrices, and introduces a world matrix resolver that runs on trait updates. This has the following benefits:

* Matrices over poses makes both world transforms and edited / live transforms much simpler and performant to calculate
* Three.js objects are no longer queried on the fly to obtain world position and rotation
* Matrix updates for Three.js objects now only happen on network response, not every frame (the default in Three.js)
* We can now move to instanced rendering of all objects based on WorldMatrix, solving transparency and performance issues
* We can more easily render based on trait since the Entity now is the source of truth for all position and rotation data

### Followups

After this, the remaining refactor steps are:
* Remove `<Portal>` and render all objects based on `MatrixWorld`. 
* Instance render all geometries.


src/lib/ecs/useTrait.svelte.ts
```diff
@@ -33,32 +29,39 @@ export function isWorld(target: Entity | World | null | undefined): target is Wo
...
-): { current: TraitRecord<T> | undefined } {
+): { readonly current: TraitRecord<T> | undefined } {
 	const contextWorld = useWorld()
-	const targetEntity = $derived(target())
-	const world = $derived(isWorld(targetEntity) ? targetEntity : contextWorld)
-	const entity = $derived(isWorld(targetEntity) ? targetEntity[internal].worldEntity : targetEntity)
-
-	// Initialize the state with the current value of the trait.
-	let value = $derived(entity?.get(trait))
+	let value = $state.raw<TraitRecord<T>>()
```
micheal-parks, line 34: Reactivity bugfix, will need to add it to my upstream koota PR

src/lib/components/overlay/Details.svelte
```diff
@@ -236,65 +247,28 @@
 		}
 	}
 
-	useTask(
```
micheal-parks, line 239: 🥳

mattmacf98 (approved):
approved with comments to:
1. double check microtask is 1-1 with animation frame
2. move transform math stuff to lib/three (to be renamed lib/math)

DTCurrie (approved):

PR #667, opened 2026-05-08
Fix: clicking on frames in the Treeview causes reparenting due to Tweakpane behavior  

micheal-parks:


src/lib/components/Scene.svelte
```diff
@@ -29,10 +29,14 @@
 
 	let { children }: Props = $props()
 
+	const threlte = useThrelte()
 	const settings = useSettings()
 	const focusedObject3d = useFocusedObject3d()
 	const origin = useOrigin()
 
+	// @ts-expect-error This is for debugging
+	globalThis.__threlte__ = threlte
```
DTCurrie, line 38: do we want to keep this?

src/lib/ecs/useWorld.ts
```diff
@@ -1,11 +1,21 @@
...
 
 export function provideWorld() {
 	const world = createWorld()
 
+	// @ts-expect-error This is for debugging.
+	globalThis.__koota__ = {
+		world,
+		traits,
+		relations,
+	}
```
DTCurrie, line 17: and this?

DTCurrie (approved):

src/lib/components/Scene.svelte
micheal-parks: I'd vote for it. I don't see any danger, and it saves me adding it and un-adding it all the time

PR #674, opened 2026-05-12
add generic component to useGeometries

gmulz:
merged a change to rdk that adds GetGeometries to generic components (seems like a bug that it was not implemented) and now closing the loop so the 3d scene tab can visualize generic components with geometries. should let us do very simple things like make obstacles out of meshes that aren't implemented as "grippers" for no reason

tested this locally and it does work with the rdk changes (i forgot to take a screenshot before shutting everything off but trust me it works)

rdk pr: https://github.com/viamrobotics/rdk/pull/5999

### DIFF
--- a/src/lib/hooks/useGeometries.svelte.ts
+++ b/src/lib/hooks/useGeometries.svelte.ts
@@ -1,4 +1,11 @@
-import { ArmClient, BaseClient, CameraClient, GantryClient, GripperClient } from '@viamrobotics/sdk'
+import {
+	ArmClient,
+	BaseClient,
+	CameraClient,
+	GantryClient,
+	GenericComponentClient,
+	GripperClient,
+} from '@viamrobotics/sdk'
 import {
 	createResourceClient,
 	createResourceQuery,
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useGeometries.svelte.ts
```diff
@@ -37,6 +44,7 @@ export const provideGeometries = (partID: () => string) => {
 	const cameras = useResourceNames(partID, 'camera')
 	const grippers = useResourceNames(partID, 'gripper')
 	const gantries = useResourceNames(partID, 'gantry')
+	const generics = useResourceNames(partID, 'generic')
```
DTCurrie, line 47: Any issues with this and generic services?

DTCurrie (changes requested):
don't forget to add a changeset

src/lib/hooks/useGeometries.svelte.ts
gmulz: oops yes

micheal-parks (approved):

src/lib/hooks/useGeometries.svelte.ts
gmulz: fixed, but is this a shortcoming of the svelte sdk that i can't filter by anything other than subtype?

DTCurrie: yeah we could open it up to specifying a type as well, or the whole api triple if you want. I don't know if there are other API subtype conflicts between services and components, but it should be straightforward to support.

DTCurrie (approved):

PR #676, opened 2026-05-12
[APP-16427] 3D scene reverts to old obstacle position after save before updating

micheal-parks:
Fixes temporary state when saving in app.viam.com. Also includes a small drive by fix for missing options in the parent selector dropdown.


mattmacf98 (approved):
approved pending running e2e for frame editing to confirm no regressions

DTCurrie (approved):

PR #678, opened 2026-05-14
assign 3d models current after all clients resolve

mattmacf98:
# Description 
in this thread https://viaminc.slack.com/archives/C071CRAU7EJ/p1778765771671389

we saw issue when visualizing multiple 3d models for different resources in the same scene, I was setting current in the inner loop so it would update for the first arm but just add more keys on subsequent sets (which svelte was not reactive to since it is a state.raw). This worked fine on other real arms since the geometry labels are set by the module (so base_top is always xarm:base_top for all arms) the fake and simulated change the label prefix to match the frame e.x. arm1:base_top and arm2:base_top

This PR sets current outside of the loop after the full object is accumulated

# Testing 
- tested on Brandon's example with sim arms 
Screenshot 2026-05-14 at 2 56 01 PM
- tested on a real robot with 2 arms 
Screenshot 2026-05-14 at 2 58 35 PM

### DIFF
--- a/src/lib/hooks/use3DModels.svelte.ts
+++ b/src/lib/hooks/use3DModels.svelte.ts
@@ -68,12 +68,12 @@ export const provide3DModels = (partID: () => string) => {
 						}
 					})
 				}
-				current = next
 			} catch (error) {
 				// some arms may not implement this api yet
 				console.warn(`${client.current.name} returned an error: ${error} when getting 3D models`)
 			}
 		}
+		current = next
 	}
 
 	$effect(() => {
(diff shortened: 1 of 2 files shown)


src/lib/hooks/use3DModels.svelte.ts
mattmacf98, line 48: if we get to. point where we have tons of different 3d models in the scene at once, we could parallelize these calls with promise.all, seemed pre-mature to do in this pr

micheal-parks: even better, it would be nice to know if we already fetched the arm for a matching model and just use local cache

micheal-parks (approved):

PR #683, opened 2026-05-15
Use MatrixWorld trait to transform all entities

micheal-parks:
Simple refactor based on what's now available to use. Every instance can simply be transformed by its WorldMatrix trait, removing the need for `<Portal>` to determine hierarchy.


mattmacf98 (approved):
RIP portals

PR #686, opened 2026-05-18
APP-16260 add pov panel

DTCurrie:
Screenshot 2026-05-18 at 2 17 52 PM


src/lib/components/overlay/widgets/FramePov.svelte
```diff
@@ -0,0 +1,149 @@
...
+	// flips both axes so a Three.js render matches "what a sensor at this frame
+	// would see." If empirical testing shows the view is rolled, swap to
+	// makeRotationY for an X-flip instead.
+	const VIAM_TO_THREE_CAMERA = new Matrix4().makeRotationX(Math.PI)
+
+	const namedEntities = useQuery(traits.Name)
+	const entity = $derived(namedEntities.current.find((e) => e.get(traits.Name) === frameName))
+	const worldMatrix = useTrait(() => entity, traits.WorldMatrix)
+
+	const povCamera = new PerspectiveCamera(60, 1, 0.01, 1000)
```
micheal-parks: We should have a perspective / ortho switch in this toggle too

src/lib/components/overlay/widgets/FramePov.svelte
```diff
@@ -0,0 +1,149 @@
...
+	let isOpen = $state(true)
+	let viewportEl = $state.raw<HTMLDivElement>()
+
+	const composed = new Matrix4()
+	const originMat = new Matrix4()
+	const savedViewport = new Vector4()
+	const savedScissor = new Vector4()
+
+	$effect(() => {
+		void worldMatrix.current
```
micheal-parks: This shouldn't be necessary if you merge main since worldMatrix now triggers invalidate on renderable items

src/lib/components/overlay/widgets/FramePov.svelte
```diff
@@ -0,0 +1,149 @@
...
+			// camera needs the same origin transform to share coordinate space
+			// with the meshes it's rendering.
+			originMat
+				.makeRotationZ(origin.rotation)
+				.setPosition(origin.position[0], origin.position[1], origin.position[2])
+			composed.copy(originMat).multiply(worldMat).multiply(VIAM_TO_THREE_CAMERA)
+			composed.decompose(povCamera.position, povCamera.quaternion, povCamera.scale)
+
+			povCamera.aspect = rect.width / rect.height
+			povCamera.updateProjectionMatrix()
```
micheal-parks, line 146: Why'd we opt out of `<View>`? It would eliminate pretty much all the code in this file minus setting z up and managing open / closed state

DTCurrie: The view creates its own scene context, so we would have to remount the entire scene inside of it. That felt sketchy and was difficult to maintain. I could be wrong about that, but as I was piecing it together that's how it looked.

micheal-parks: Agh, totally forgot about that design element. Hmmm, yeah, good call out. Totally makes sense to go with this but definitely exposes a major use case gap in `<View>` that I'm going to think about. I think it needs an optional scene prop:

`<View {scene} />`

that it would just then use instead of creating a context. That way you get multi-perspective rendering without scene replication

DTCurrie: Yeah that makes sense.

micheal-parks (approved):

PR #688, opened 2026-05-19
Fix parenting on `GetGeometries` entities

micheal-parks:
### Overview

We were most of the way there to supporting parenting to `Geometry` entities. The `Center` prop not informing local matrices was the missing step. This PR fixes that, with a few other drive by memory leak fixes.

### DIFF
--- a/src/lib/components/overlay/Details.svelte
+++ b/src/lib/components/overlay/Details.svelte
@@ -2,7 +2,6 @@
 	module
 	lang="ts"
 >
-	import { ThemeUtils } from 'svelte-tweakpane-ui'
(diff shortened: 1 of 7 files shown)


src/lib/components/overlay/Details.svelte
```diff
@@ -280,12 +279,6 @@
 			2
 		)
 	}
-
-	ThemeUtils.setGlobalDefaultTheme({
```
micheal-parks, line 284: This was in a bad spot, moved it up to app root

DTCurrie (approved):

PR #693, opened 2026-05-19
add shapes overlay

DTCurrie:
Note: Quickly put this together to support the education training effort. Open to suggestions/commits on this. It replaces the old `=` hotkey.

Adds an "Add Shape" dashboard button to allow creating simple geometries. Also adds color and visibility controls to the details panel.

As a follow-up, I am going to look into splitting up the detail panel into tabs, it is getting overloaded and we will want space for controls soon.

Screenshot 2026-05-20 at 10 19 44 AM
Screenshot 2026-05-20 at 10 19 57 AM
Screenshot 2026-05-20 at 10 20 09 AM
Screenshot 2026-05-20 at 10 20 24 AM


src/lib/components/overlay/Details.svelte
```diff
@@ -697,8 +777,46 @@
...
+							value={colorValue}
+							type="float"
+							on:change={handleColorChange}
+						/>
+					</div>
+				</div>
+			{/if}
+
+			<div aria-label="mutable visibility">
+				<Checkbox
```
micheal-parks, line 813: Why are we adding a checkbox for invisibility? Doesn't this duplicate the Treeview invisibility button?

src/lib/components/InputBindings.svelte
```diff
@@ -156,4 +156,10 @@
 			entity?.add(traits.Invisible)
 		}
 	})
+
+	keys.onKeys('-', () => {
```
micheal-parks, line 160: I'd maybe leave this out until we have undo keybindings + modal, the trash button works and this could cause confusion

src/lib/FrameConfigUpdater.svelte.ts
```diff
@@ -9,6 +9,23 @@ import { createPose, matrixToPose, poseToMatrix } from '$lib/transform'
 
 const tempPose = createPose()
 
+/**
+ * Picks the matrix trait to edit. Config-backed frames (with FramesAPI) stage
+ * changes in EditedMatrix and round-trip through the robot config; local
+ * Removable entities write directly to Matrix and skip the config call.
+ */
+const getEditTarget = (entity: Entity) => {
```
micheal-parks, line 17: These file edits concern me a lot. I don't like how entangled this is getting. I'm working on a PR that extracts all of frame editing into a plugin, and I was going to have completely separate pathway for editing local entities. Coupling them like this seems very risky for stability.

src/lib/components/Entities/Mesh.svelte
```diff
@@ -49,6 +51,7 @@
 	const box = useTrait(() => entity, traits.Box)
 	const capsule = useTrait(() => entity, traits.Capsule)
 	const sphere = useTrait(() => entity, traits.Sphere)
+	const plane = useTrait(() => entity, traits.Plane)
```
mattmacf98, line 54: nit: a little iffy about adding planes as a geo type since it is not a viam supported one and people may be confused why they can set geo to a plane for this flow but not for frame geos

src/lib/components/Entities/Mesh.svelte
```diff
@@ -137,9 +143,13 @@
 		renderOrder={renderOrder.current}
 		{...rest}
 	>
-		{#if box.current || sphere.current}
-			{@const meshGeometry = box.current ? unitBox : unitSphere}
-			{@const edgesGeometry = box.current ? unitBoxEdges : unitSphereEdges}
+		{#if box.current || sphere.current || plane.current}
+			{@const meshGeometry = box.current ? unitBox : sphere.current ? unitSphere : unitPlane}
+			{@const edgesGeometry = box.current
```
mattmacf98, line 148: nit: is this enough nested ternaries to warrant a helper func?

src/lib/components/overlay/AddShape.svelte
```diff
@@ -0,0 +1,111 @@
...
+
+	const kinds: { value: ShapeKind; label: string; icon: unknown }[] = [
+		{ value: 'box', label: 'Box', icon: Box },
+		{ value: 'sphere', label: 'Sphere', icon: Circle },
+		{ value: 'capsule', label: 'Capsule', icon: Pill },
+		{ value: 'plane', label: 'Plane', icon: Square },
+		{ value: 'frame', label: 'Frame', icon: Axis3d },
+	]
+
+	let counters = $state<Record<ShapeKind, number>>({
```
mattmacf98, line 32: do we decrement this on remove? if not maybe we should rename this to shapeIds so people don't mistake this as a true count of the number of them in the scene

src/lib/components/overlay/AddShape.svelte
```diff
@@ -0,0 +1,111 @@
...
+	let name = $derived(`${kind} ${counters[kind] + 1}`)
+
+	const isValid = $derived(name.trim().length > 0)
+
+	const handleAdd = () => {
+		const trimmed = name.trim()
+		if (!trimmed) return
+
+		const entity = spawners[kind](world)
+		entity.set(traits.Name, trimmed)
```
mattmacf98, line 53: is this not already handled in the baseTraits in the spawner?

```
const baseTraits = (kind: ShapeKind) => [
	traits.Name(`custom ${kind} ${++counters[kind]}`),
	traits.Matrix,
	traits.Removable,
	traits.Transformable,
]
```

could the source of truth ids just be in there instead?

src/lib/FrameConfigUpdater.svelte.ts
```diff
@@ -9,6 +9,23 @@ import { createPose, matrixToPose, poseToMatrix } from '$lib/transform'
 
 const tempPose = createPose()
 
+/**
+ * Picks the matrix trait to edit. Config-backed frames (with FramesAPI) stage
```
mattmacf98, line 13: I think this expands the responsibility of the FrameConfigUpdater.svelte.ts‎ too much (it now deals with frame updates and the custom geo ones)

I think it would be good to create a new updater to just handle the custom geo stuff and use that when the focused entity is a custom geo and then we can leave this file alone to just deal with frames

src/lib/components/overlay/Details.svelte
DTCurrie: consistency with other metadata rendering values, and just discoverability.

src/lib/components/overlay/AddShape.svelte
DTCurrie: not sure it matters too much, but we can. its just to avoid name collisions since these should just be temporary entities. I don't if poeple will care to much since they aren't persisted

src/lib/components/InputBindings.svelte
DTCurrie: we can, this was just to not lose the existing behavior for custom geos that were added with the =/+ key

src/lib/FrameConfigUpdater.svelte.ts
DTCurrie: oh yeah, and to @mattmacf98 comment this was just a quick way to get this stubbed in and working based off the old geos, but I think they should be separate

src/lib/components/Entities/Mesh.svelte
DTCurrie: yeah as mentioned I would probably separate this from the frame editing stuff in general this was just a quick way to get something useable to test out.

src/lib/components/overlay/Details.svelte
micheal-parks: Is that really a problem though? I worry that we're inventing the problem here - no one that I've talked to has reported the eye buttons on the tree to not be discoverable

src/lib/components/InputBindings.svelte
micheal-parks: Yeah, but this expands niche behavior to something that encompasses _all_ delete-ables

src/lib/components/overlay/Details.svelte
DTCurrie: I remember some people not knowing you could do that. Also don’t see the harm here. I actually find it useful. When I am observing a scene I will often move the world tree out of the way or shrink it to get more real estate. And if I am tweaking rendering stuff it’s weird I could do all of that in the details panel but for visibility I have to go to an entire separate panel for this one specific instance.

micheal-parks: That's fair, not opposed, just wanted to drill a bit first 👷

PR #695, opened 2026-05-20
Move Draw Service to `/plugins`

micheal-parks:
Creates a new `/plugins` export and adds `<DrawService>`. Embedded environments like app.viam.com will no longer create drawing service providers.


src/lib/components/overlay/AddRelationship.svelte
```diff
@@ -55,24 +53,12 @@
...
-			)
-		} else {
-			entity.add(
-				relations.SubEntityLink(selectedEntity, {
-					indexMapping: relationshipFormula,
-					type: linkType,
-				})
-			)
-		}
+		entity.add(
```
micheal-parks, line 56: We no longer write back relationships to the Server. It can still serve relationships and modify them, and avoid touching UI generated relationships, but this is all client-side

src/lib/components/Scene.svelte
```diff
@@ -12,9 +12,9 @@
 	import Selected from '$lib/components/Selected.svelte'
 	import SelectedTransformControls from '$lib/components/SelectedTransformControls.svelte'
 	import StaticGeometries from '$lib/components/StaticGeometries.svelte'
+	import { bvh } from '$lib/hooks/plugins/bvh.svelte'
```
micheal-parks, line 15: Threlte plugins are subsets of hooks, so figured this naming makes sense

src/lib/components/overlay/AddRelationship.svelte
DTCurrie: This was a specific feature we included. Are we just cutting it entirely?

micheal-parks: Everything should behave mostly the same - so I'd call this only slightly modifying the behavior with the gain of clear plugin isolation and one way data flow. 

Relationships that are drawing generated still go through to the client, and if the user adds new relationships in the UI and then draws more things, those new relationships shouldn't be touched by drawing updates since there's now a hook to manage relationship resolution.

DTCurrie (approved):

PR #709, opened 2026-05-26
make devtools a plugin

DTCurrie:
this _should_ fix the auto-bump PR flow. issue stems from us using tanstack v6 which is causing dependency issues in app which is still on v5 and also installs the svelte devtools package. this felt like a reasonable fix. I am going to work on updating the app repo to tanstack v6 next.


src/lib/components/App.svelte
```diff
@@ -111,8 +112,12 @@
 	})
 </script>
 
-{#if settings.current.enableQueryDevtools}
-	<SvelteQueryDevtools initialIsOpen />
+{#if devtools.isAvailable && settings.current.enableQueryDevtools}
```
micheal-parks: Why not make `<Devtools>` just a plugin an import it in the routes section? I feel like this plugin will be used less than most others, especially once we get tanstack on the same version in app

DTCurrie: I considered that, and am open to it. This was sort of the "smallest lift". But if we don't think it is "core" behavior and something that should be a plugin it should be an easy lift and shift.

DTCurrie: The follow-up would be should we also move the stats behavior to the same plugin and just call it a "Debug" plugin?

micheal-parks: Yes, totally agreed there, and totally up for just following up with that

micheal-parks (approved):

PR #715, opened 2026-05-27
disallow edit frames if component comes from fragment with varibles d…

mattmacf98:
# Details 
In this ticket https://viam.atlassian.net/browse/APP-15772 we saw people had issues editing frames coming from fragments that use variables overwriting the variables. This is technically correct (fragment mods should override variables)

```
{
  "fragments": [
    {
      "id": "a8c0baf1-8e14-4bae-aa74-3c98b172262a",
      "variables": {
        "translation_x": 1000
      }
    }
  ],
  "fragment_mods": [
    {
      "fragment_id": "a8c0baf1-8e14-4bae-aa74-3c98b172262a",
      "mods": [
        {
          "$set": {
            "components.base-1.frame": {
              "translation": {
                "x": 100,
                "y": 0,
                "z": 0
              },
              "parent": "world",
              "orientation": {
                "type": "ov_degrees",
                "value": {
                  "x": 0,
                  "y": 0,
                  "z": 1,
                  "th": 0
                }
              },
              "geometry": {
                "type": "box",
                "x": 100,
                "y": 100,
                "z": 100
              }
            }
          }
        }
      ]
    }
  ]
}
```

this was a little confusing to users who didn't understand they were overwriting variables so we will disable editing components from fragments which have variables

# Testing
- ran locally, saw I can still edit components from fragments with no variables and saw that I can no longer edit when they have variables

https://github.com/user-attachments/assets/f5fcbfe1-d773-4789-91be-9b3bea1cdd72

Screenshot 2026-05-27 at 3 05 25 PM


src/lib/hooks/usePartConfig.svelte.ts
```diff
@@ -301,7 +306,7 @@ export const usePartConfig = (): PartConfigContext => {
 interface AppEmbeddedPartConfigProps {
 	current: Struct
 	isDirty: boolean
-	componentToFragId: Record<string, string>
+	componentToFragInfo: Record<string, FragmentInfo>
```
mattmacf98: this will require a change when we bump motion-tools in app to change the param type

src/lib/components/overlay/Details.svelte
```diff
@@ -109,7 +109,13 @@
 
 	const isFrameNode = $derived(!!framesAPI.current)
 	const isGeometry = $derived(!!geometriesAPI.current)
-	const showEditFrameOptions = $derived(isFrameNode && partConfig.hasEditPermissions)
+	const isFragmentComponentWithVariables = $derived(
```
micheal-parks, line 112: Maybe we should add a small message in the details panels saying something along the lines of "This frame is not editable due to the presence of fragment variables"?

micheal-parks (approved):

src/lib/components/overlay/Details.svelte
mattmacf98: oh yeah I'll add something for that

DTCurrie (approved):

PR #716, opened 2026-05-27
select plugin emit multiple entities per selection

mattmacf98:
## Stack
---here---> 1. https://github.com/viamrobotics/visualization/pull/716
2. https://github.com/viamrobotics/visualization/pull/727

# Description 
This PR changes behavior of the select tool to allow emitting multiple entities for a given selection (if a selection touches multiple source entities). This is needed for a SOW web app request to be able to invalidate selections which touch multiple zones (represented by multiple PCDs)

# Testing
- tested in the /selection playground url

https://github.com/user-attachments/assets/875a30a8-c872-4b8e-9747-44287b88f352

PR #727, opened 2026-05-29
export snapshot pcd helper

mattmacf98:
## Stack
1. https://github.com/viamrobotics/visualization/pull/716
---here---> 2. https://github.com/viamrobotics/visualization/pull/727

# Description
This PR adds a helper to get PCD info from a snapshot to help with sanding use case of sending target pcds in a snapshot instead of as an array

# Testing
- confirmed working here https://github.com/viamrobotics/sanding-webapp/pull/145


src/lib/snapshot.ts
```diff
@@ -205,3 +206,34 @@ const getRenderArmModels = (
...
+	
+/**
+ * Decodes every point-cloud drawing in a snapshot into `{ name, positions, colors }`,
+ * keyed and merged by `referenceFrame`. Chunked clouds share a referenceFrame, so their
+ * positions/colors are concatenated back into one cloud. Inverse of `DrawPoints`.
+ *
+ * Colors are gated through `isVertexColors`: only genuine per-vertex color arrays are kept;
+ * single-uniform-color arrays (3 bytes) are discarded as `undefined`.
+ */
+export const decodeDrawnSnapshotPointClouds = (snapshot: Snapshot): SnapshotPointCloud[] => {
```
mattmacf98, line 224: would be nice to at some point make this more generic to extract data from all drawings AND transforms for client use when they have a snapshot, will need to think more about the types interface though and how we extract them more

DTCurrie: some to rethink with the v2 move from snapshots -> scenes

DTCurrie (approved):

PR #728, opened 2026-05-29
Add multiselect / multifocus

micheal-parks:
### Overview

This PR adds multiselect / multifocus support by switching us over from custom hooks to ECS traits, specifically `traits.Selected`. Focus view is also reworked: there's no more cloning objects and switching scene content, it now accepts all selected items and just makes all others invisible, plus a few other scene tweaks that were pre-existing. This results in a much more efficient swap between focused / unfocused mode, especially in the case of pointclouds.

For now details panels just stack during multiselect (shift + click). I figured this was fine, the benefits of multiselect are strong enough to make this UX better later.

multiselect

### Changelog 

* Add `Selected` trait and allow tagging any resource with it. 
* Make focus view accept all selected items
* Fix Treeview so that expanding only happens on chevron and not text click (to match our other treeviews elsewhere)


src/lib/components/Entities/Arrows/Arrows.svelte
```diff
@@ -20,19 +19,17 @@
 
 	const { invalidate } = useThrelte()
 	const worldMatrix = useTrait(() => entity, traits.WorldMatrix)
+	const selected = useTag(() => entity, traits.Selected)
 	const invisible = useTrait(() => entity, traits.InheritedInvisible)
 	const showAxesHelper = useTrait(() => entity, traits.ShowAxesHelper)
 
 	const events = useEntityEvents(() => entity)
-	const selectedEntity = useSelectedEntity()
-	const focusedEntity = useFocusedEntity()
```
micheal-parks, line 28: We're able to simplify a lot of components now that focusing is just a flipped state and not attached to any entity

src/lib/components/Entities/hooks/useEntityEvents.svelte.ts
```diff
@@ -98,15 +96,6 @@ export const useEntityEvents = (entity: () => Entity | undefined) => {
 		}
 	}
 
-	const ondblclick = (event: IntersectionEvent<MouseEvent>) => {
```
mattmacf98, line 101: nit: fine to remove, but we could add a focused trait maybe and make it a singleton so it clears all other focused trait on entities before setting

src/lib/components/KeyboardBindings.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
```
micheal-parks, line 1: These need to be separated out: we switch camera controls with trackball controls but want to still keep universal keybindings

src/lib/components/overlay/dashboard/Dashboard.svelte
```diff
@@ -71,6 +86,18 @@
 		/>
 	</fieldset>
 
+	<!-- focus -->
+	<fieldset class="flex">
```
mattmacf98: sanding won't be super happy about another dashboard button

src/lib/components/App.svelte
```diff
@@ -107,9 +107,15 @@
 		environment.current.isStandalone = !localConfigProps
 	})
 
+	let installed = false
 	$effect(() => {
+		if (installed) return
```
mattmacf98: a regular onMount or effect with no deps isnt enough to do a single invoke?

src/lib/plugins/Selection/SelectionTool.svelte
```diff
@@ -58,7 +58,8 @@
 
 		const newest = newEntities.at(-1)
 		if (newest === undefined) return
-		selectedEntity.set(newest)
+
```
mattmacf98, line 62: this should also clear the other selection pcd selection entities to stay 1:1 (eventually we should just move this all to SOW app)

src/lib/components/overlay/dashboard/Dashboard.svelte
micheal-parks: Yeah very agreed I'm thinking about pluginizing this before merge

src/lib/components/App.svelte
micheal-parks: yeah can switch to onmount

src/lib/components/Entities/hooks/useEntityEvents.svelte.ts
micheal-parks: what would be the purpose of the focus trait? it sounds like doing this would break multi-object focus?

src/lib/components/overlay/dashboard/Dashboard.svelte
micheal-parks: pluginized 🧩

src/lib/components/App.svelte
micheal-parks: Oh, I misread. Changed this to an onMount, but to answer your question, something was causing this effect to re-run, not sure though

src/lib/plugins/Selection/SelectionTool.svelte
micheal-parks: done

src/lib/plugins/MeasureTool/MeasureTool.svelte
```diff
@@ -73,9 +71,9 @@
 	}
 
 	$effect(() => {
-		// eslint-disable-next-line @typescript-eslint/no-unused-expressions
-		;(focusedEntity.current, enabled)
-		untrack(() => clear())
+		if (!enabled) {
+			untrack(() => clear())
+		}
```
DTCurrie, line 76: nit:
```suggestion
		if (!enabled) untrack(() => clear())
```

mattmacf98 (approved):

DTCurrie (approved):

PR #735, opened 2026-06-01
app-16292: shared ECS traits and helpers for entity pose plugins

DTCurrie:
Adds shared ECS traits and helpers (`writeMatrix`, `CustomDetails`), a `'gizmo'` interaction mode, and a `firstHitOnly` raycaster option. Foundation for the Gizmos plugin in #738. No consumers in this PR.

> [!NOTE]

### Stack

1. **This PR**: Shared ECS traits and helpers
2. Details refactor (#736)
3. Component tweaks (#737)
4. Gizmos plugin (#738)
5. Gizmos plugin docs (#739)

### Frontend

- `src/lib/ecs/traits.ts`: adds `writeMatrix(entity, patch)` that mutates an entity's `Matrix` trait via the existing `createPose` / `matrixToPose` / `poseToMatrix` round-trip and then dirties the trait.
- `src/lib/ecs/traits.ts`: adds the `CustomDetails` marker trait so the refactored Details panel can suppress its default frame and pose blocks for entities that render their own. May be removed after frame editing becomes a plugin.
- `src/lib/hooks/useMouseRaycaster.svelte.ts`: accepts a `firstHitOnly` option that is forwarded to `three-mesh-bvh`, for tools that only care about the closest hit.
- `src/lib/hooks/useSettings.svelte.ts`: extends the `Settings.interactionMode` union with `'gizmo'`.

### Testing

Added `src/lib/ecs/__tests__/writeMatrix.spec.ts` covering each `Pose` field individually and the no-op path when the entity has no `Matrix` trait. Ran `pnpm check` and `pnpm test`.


src/lib/ecs/traits.ts
```diff
@@ -74,6 +75,14 @@ export const InstancedMatrix = trait(() => ({
 export const Hovered = trait(() => true)
 export const Invisible = trait(() => true)
 
+/**
+ * Suppresses the default frame-style world/local pose and parent-frame blocks
+ * in the details panel. Entities that render their own pose UI via the
+ * `details-extensions` portal target (e.g. gizmo plugin entities) opt in by
+ * adding this trait.
+ */
+export const CustomDetails = trait(() => true)
```
DTCurrie, line 84: This can probably be removed when frame editing is made a plugin

src/lib/hooks/useMouseRaycaster.svelte.ts
```diff
@@ -110,6 +116,11 @@ export const useMouseRaycaster = (getOptions?: () => { enabled: boolean }) => {
 			return
 		}
 
+		// firstHitOnly is set on the shared raycaster instance, so it applies to
+		// both onPointerMove (hover) and onPointerUp (click). This is intentional
+		// for gizmo-style consumers that only care about the closest hit.
+		raycaster.firstHitOnly = options.firstHitOnly
```
DTCurrie: we are using this to detect surface hits for placing gizmos. there may be a better way to handle this.

src/lib/hooks/useMouseRaycaster.svelte.ts
```diff
@@ -105,6 +111,15 @@ export const useMouseRaycaster = (getOptions?: () => { enabled: boolean }) => {
 		intersections = currentIntersections
 	}
 
+	// firstHitOnly is set on the shared raycaster instance, so it applies to
+	// both onPointerMove (hover) and onPointerUp (click). This is intentional
+	// for gizmo-style consumers that only care about the closest hit.
+	// Kept in its own effect so toggling firstHitOnly doesn't tear down and
+	// re-attach the event listeners.
+	$effect(() => {
+		raycaster.firstHitOnly = options.firstHitOnly
```
micheal-parks: I don't think there's any use case for `firstHitOnly: false` anywhere in the app - nor can i anticipate one, if it's currently false then that's a misconfig - just for more simplicity can we just reduce this to a single line `raycaster.firstHitOnly = true?

micheal-parks (approved):

PR #736, opened 2026-06-01
app-16292: refactor Details panel into focused subcomponents

DTCurrie:
Splits the monolithic `Details.svelte` overlay into focused subcomponents under `src/lib/components/overlay/details/`. Behavior is unchanged for the refactor itself.

This is preparation for the per-entity panels that the Gizmos plugin mounts in #738.

> [!NOTE]

### Stack

1. Shared ECS traits and helpers (#735)
2. **This PR**: Details refactor
3. Component tweaks (#737)
4. Gizmos plugin (#738)
5. Gizmos plugin docs (#739)

### Frontend

- Added `src/lib/components/overlay/details/ColorDetails.svelte`, `GeometryDetails.svelte`, `MatrixDetails.svelte`, `LineDetails/LineDetails.svelte`, `OpacityDetails.svelte`, and `AxesHelperDetails.svelte`.
- `src/lib/components/overlay/details/LineDetails/linePositions.ts`: extracts the line-position derivation with its own spec so it can be tested without rendering Svelte.
- `src/lib/components/overlay/Details.svelte`: mounts a `PortalTarget` named `details-extensions` so plugin-supplied editors (e.g. the gizmo Details in #738) can render inline, and replaces the inline opacity slider and show-axes-helper switch with `<OpacityDetails>` and `<AxesHelperDetails>`. Plugin-supplied panels mount these conditionally per entity kind.
- `MatrixDetails.svelte`: takes required `onPoseChange(patch)` and `onParentChange(next)` props so the caller decides whether mutations commit immediately (`writeMatrix` / `hierarchy.setParent`) or buffer through an edit session (e.g. a future frame edit plugin writing to `EditedMatrix`). Required rather than defaulted to avoid silently committing edits that should have been buffered.
- `OpacityDetails.svelte`: owns the opacity slider. Always writes `traits.Opacity(value)` instead of removing the trait at `1`, and falls back to `0.7` for display to match the renderer's historical default. Previously, the slider would remove the trait at `1`, after which `Line` / `Mesh` would fall back to their own `0.7` default and render translucent instead of opaque.
- `AxesHelperDetails.svelte`: owns the show-axes-helper switch, adding/removing `traits.ShowAxesHelper` directly.
- `vitest-setup-client.ts`: stubs `@threlte/extras` `Portal`, `PortalTarget`, and `HTML` so the refactored panel can render in unit tests without a `<Canvas>` parent.

### Testing

The existing `src/lib/components/overlay/__tests__/Details.svelte.spec.ts` continues to pass against the refactored component with the new mocks. New `MatrixDetails.svelte.spec.ts`, `OpacityDetails.svelte.spec.ts`, and `AxesHelperDetails.svelte.spec.ts` cover the structural rendering branches of each extracted component, and `linePositions.spec.ts` covers the extracted line-position derivation. Ran `pnpm check` and `pnpm test`. Manually verified the opacity slider now produces a fully opaque mesh at `1` and stays at `0.7` for a trait-less entity.


micheal-parks (commented):
should we postfix all of these new files with "Details" if they're in the details folder? seems potentially noisy like the situation with "BaseView", "CameraView" etc in test-cards, but I could also understand not wanting to pollute ctrl+p search

src/lib/components/overlay/Details.svelte
```diff
@@ -82,9 +85,10 @@
 	const removable = useTrait(() => entity, traits.Removable)
 	const points = useTrait(() => entity, traits.Points)
 	const arrows = useTrait(() => entity, traits.Arrows)
-	const opacity = useTrait(() => entity, traits.Opacity)
 	const framesAPI = useTrait(() => entity, traits.FramesAPI)
 	const geometriesAPI = useTrait(() => entity, traits.GeometriesAPI)
+	const customDetails = useTrait(() => entity, traits.CustomDetails)
```
micheal-parks: We can use `useTag` now to avoid an extra line below

src/lib/components/overlay/Details.svelte
```diff
@@ -146,25 +154,42 @@
...
+
 	const stopKeyboardPropagation = (event: KeyboardEvent) => {
 		event.stopPropagation()
 	}
 
 	const handlePositionChange = (event: PointChangeEvent) => {
 		if (event.detail.origin !== 'internal' || !entity) return
 		const next = event.detail.value as PointValue3dObject
-		detailConfigUpdater.updateLocalPosition(entity, next)
+		if (isFrameNode) {
```
micheal-parks: reminder to rip this out

src/lib/components/overlay/details/PoseDetails.svelte
```diff
@@ -0,0 +1,189 @@
+<script
```
micheal-parks, line 1: Can we call this PoseDetails? I feel like this should reflect what the user sees, matrices are an internal design detail that the user never sees

DTCurrie: Ha, I actually originally named it that then swapped to `MatrixDetails`. I'll swap it back

micheal-parks (approved):

PR #738, opened 2026-06-01
app-16292: add Gizmos plugin

DTCurrie:
Adds the `Gizmos` plugin: a dashboard toolset for placing editable scene aids (coordinate systems, reference planes, reference geometries, polylines, arrows, vertex normals, surface normals) by clicking on existing surfaces. Each placed entity is auto-selected so its plugin-supplied Details panel opens for in-place editing. Stacks on #737.

> [!NOTE]

### Stack

1. Shared ECS traits and helpers (#735)
2. Details refactor (#736)
3. Component tweaks (#737)
4. **This PR**: Gizmos plugin
5. Gizmos plugin docs (#739)

### Frontend

- `src/lib/plugins/Gizmos/Gizmos.svelte`: plugin entry. Mounts a dashboard popover, a Threlte portal that hosts the active tool, and per-entity renderers via `GizmoEntities.svelte`. Switches the visualizer into `'gizmo'` interaction mode while a tool is armed and reverts to `'navigate'` on exit.
- `src/lib/plugins/Gizmos/tools/` (`ArrowTool`, `CoordinateSystemTool`, `GeometryTool`, `LineTool`, `NormalsTool`, `PlaneTool`): one component per gizmo kind. Each delegates the click / drag pipeline to the shared `usePlace`, `useInputs`, and `usePending` hooks.
- `src/lib/plugins/Gizmos/GizmoArrow.svelte`, `GizmoPlane.svelte`, `GizmoPolylineMeasure.svelte`, `GizmoNormals.svelte`, `GizmoDetails.svelte`: per-entity renderers and Details editors. They carry the `CustomDetails` trait from #735 so the default frame and pose blocks in the Details panel are suppressed.
- `GizmoDetails.svelte`: renders the shared `MatrixDetails` from #736 with `onPoseChange={traits.writeMatrix}` and `onParentChange={hierarchy.setParent}`. Gizmos commit pose / parent edits immediately; a future frame edit plugin would pass session-buffered handlers instead.
- `src/lib/plugins/Gizmos/useGizmos.svelte.ts`: provider plus `useGizmos()` hook exposing the active `mode`, per-tool options (plane axis / construction / offset, geometry shape / construction / wireframe, line space / measure, arrow axis, vertex and surface normals length), and an `exit()` action.
- `src/lib/plugins/Gizmos/traits.ts`: marker traits `Gizmo`, `PendingGizmo`, `ReferencePlane`, `GizmoArrow`, `VertexNormals`, `SurfaceNormals`, and `PolylineMeasure` for queries against placed gizmos.
- Pure helpers `cursor.ts`, `surface.ts`, `matrix.ts`, `spawn.ts`, `buildNormalsMesh.ts`, and `buildSurfaceNormalsGeometry.ts` each have a spec under `__tests__/`.
- `src/lib/plugins/index.ts`: re-exports `Gizmos`, `gizmoTraits`, and `useGizmos` from the plugin barrel.
- `src/routes/+layout.svelte`: mounts `<Gizmos />` in the demo route so the dev app exercises the plugin.

### Testing

New specs under `src/lib/plugins/Gizmos/__tests__/` cover the pure helpers (`buildNormalsMesh`, `buildSurfaceNormalsGeometry`, `cursor`, `matrix`, `spawn`, `surface`). The full vitest suite passes with `pnpm test`. `pnpm check` is clean.

Manual smoke test plan: exercise each tool from the dashboard popover (coordinate system; reference plane in free and offset modes; reference geometry box / sphere / capsule, solid and wireframe, at-origin and free; polyline in world and screen space with each measure mode; arrow on each axis and on the surface normal; vertex normals; surface normals). Confirm `Esc` and clicking the dashboard toggle exit gizmo mode.


src/lib/plugins/Gizmos/traits.ts
```diff
@@ -0,0 +1,24 @@
...
+export type PolylineMeasureMode = 'segment' | 'total'
+
+/**  Marks an entity as a user-placed gizmo, not part of the robot's frame system.*/
+export const Gizmo = trait(() => true)
+
+/** Transient tag present on a gizmo while the user is still placing or orienting it. */
+export const PendingGizmo = trait(() => true)
+
+/** Renderable plane gizmo; normal is +Z in the entity's local frame. */
+export const ReferencePlane = trait({ width: 500, height: 500 })
```
micheal-parks:
```suggestion
export const Plane = trait({ width: 500, height: 500 })
```
I'd maybe put this with the rest of the geometry? when looking at the props it seems weird to load it with use case

src/lib/plugins/Gizmos/traits.ts
```diff
@@ -0,0 +1,24 @@
+import { trait } from 'koota'
+
+export type PolylineMeasureMode = 'segment' | 'total'
+
+/**  Marks an entity as a user-placed gizmo, not part of the robot's frame system.*/
+export const Gizmo = trait(() => true)
```
micheal-parks: We can use tags now:

```suggestion
export const Gizmo = trait()
```

src/lib/plugins/Gizmos/GizmoNormals.svelte
```diff
@@ -0,0 +1,145 @@
...
+	const sourceBuffer = useTrait(() => source, traits.BufferGeometry)
+
+	let helper = $state.raw<LineSegments | VertexNormalsHelper>()
+	let sourceMesh: Mesh | undefined
+
+	const length = $derived(
+		(kind === 'surface' ? surfaceConfig.current?.length : vertexConfig.current?.length) ?? 100
+	)
+
+	$effect.pre(() => {
```
micheal-parks: Looks like these could all be $effect since none of them have downstream effects that read from them?

src/lib/plugins/Gizmos/GizmoPlane.svelte
```diff
@@ -0,0 +1,102 @@
...
+	const invisible = useTrait(() => entity, traits.InheritedInvisible)
+	const events = useEntityEvents(() => entity)
+
+	const mesh = new Mesh()
+	const group = new Group()
+	group.matrixAutoUpdate = false
+
+	const currentOpacity = $derived(opacity.current ?? 1)
+
+	$effect.pre(() => {
```
micheal-parks: same here: $effect.pre -> $effect

src/lib/plugins/Gizmos/traits.ts
DTCurrie: I wanted to really make sure it was obvious this was not a viam-supported geometry type, thus the separation. they also have some additional options that it didn't want to have to thread into the geometry tool and make it more complicated.

src/lib/plugins/Gizmos/traits.ts
DTCurrie: Ah yeah I updated that for the custom details one but not the others, will do that.

src/lib/plugins/Gizmos/GizmoNormals.svelte
DTCurrie: Ah yeah, I think claude added these back in for "consistency". so we may need to tweak its rules a bit

src/lib/plugins/Gizmos/traits.ts
micheal-parks: wouldn't the additional options be other traits though? and I get that - but that sounds like something that would be surfaced to the users, not the underlying data itself. Plane to me is more ECS-forward, it's a trait that communicates a geometry type and therefore a shape to render - traits (components) should be as stripped from complicated ideas as possible

DTCurrie: No it has specific placement options that are different than the geometries, like the plane it is on and how far to offset it on that. So it literally has different placement options in the gizmo menu than geometries. The name doesn't really matter, but it is a different user-facing experience.

I combined them into a "reference shape" menu and just dynamically swap out the options when plane is selected, made the menu more complicated but its not the end of the world unless we add another geometry type with its own options. check it out. honestly not sure if I like it, because now the plane option is buried and hard to discover. If I saw "shape" I am not sure I think "plane" because that isn't a thing in viam, so people will have to find that. I think it would be even worse if we kept the name "reference geometry"

honestly after using it, I am thinking of reverting that change and going back to "reference plane" being its own menu section.

.claude/rules/svelte.md
```diff
@@ -46,6 +46,8 @@ This project renders a 3D scene using [Threlte](https://threlte.xyz/llms-full.tx
 
 The right question to ask for when to use `$effect` vs `$effect.pre` is "does anything downstream in the same flush need to read this before render/DOM-commit?" If yes,`.pre`; if it's a pure side-effect with nothing observing the result inside the same flush, plain $effect is correct.
 
+**Do not reach for `.pre` to "avoid a one-frame render lag."** That reasoning applies to continuously-rendered scenes; it does **not** apply here. Because rendering is on-demand, a scene mutation is followed by `invalidate()`, which schedules a render for the _next_ `requestAnimationFrame`.
```
micheal-parks, line 49: This doesn't make sense for continuously rendered scenes either. `$effect.pre` runs synchronously and immediately. `$effect` runs in a batched manner in a microtask. But the guaranteed order of operations for JS per frame is:

1. Sync tasks
2. Microtasks
3. requestAnimationFrame

Microtasks always run before `requestAnimationFrame`. Therefore even in continuously running apps `$effect()` is fine. The only time it's not fine is when it's writing Three.js props that other effects will potentially read from. Then $effect.pre is more necessary to get ahead of the other reads. But this is rare and sort of dangerous.

src/lib/components/overlay/Popover.svelte
```diff
@@ -21,11 +21,20 @@
...
+	the *content* element's computed z-index. So the z-index must live on the content, not 
+	the positioner. 
+-->
 <div
 	use:portal={{ disabled: !api.portalled }}
 	{...api.getPositionerProps()}
 >
-	<div {...api.getContentProps()}>
+	<div
+		class="z-max"
```
micheal-parks, line 35: Do we need z-max? Can we use a less aggressive value? (I think we use around z-5) for other floating UI

src/lib/plugins/Gizmos/usePending.svelte.ts
```diff
@@ -0,0 +1,41 @@
...
+	const world = useWorld()
+
+	let pending = $state.raw<Entity>()
+
+	useConfirmInput(() => options().onConfirm?.())
+	useCancelInput(() => options().onCancel?.())
+	useAddNextInput(() => options().onAddNext?.())
+	useUndoInput(() => options().onUndo?.())
+
+	onDestroy(() => {
```
micheal-parks, line 27: nit: I'd do $effect(() => { return ... }) since running in a microtask is a bit safer with the reactivity runtime, and (less applicable here) onDestroy runs in SSR

src/lib/plugins/Gizmos/useInputs.svelte.ts
```diff
@@ -0,0 +1,82 @@
...
+	$effect(() => {
+		window.addEventListener('keydown', onKey)
+		return () => window.removeEventListener('keydown', onKey)
+	})
+}
+
+/** Undo the last pending tool action. */
+export const useUndoInput = (handler: () => void) => {
+	const onKey = (event: KeyboardEvent) => {
+		if (event.key !== 'Backspace') return
```
micheal-parks, line 52: Why not control + z ? I think most people would try that before backspace for undo

src/lib/plugins/Gizmos/useInputs.svelte.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Cancel a pending tool action. Escape always triggers. Right-click triggers
+ * only while a tool is active.
+ */
+export const useCancelInput = (handler: () => void) => {
+	const { dom } = useThrelte()
+	const plugin = useGizmos()
+
+	const onKey = (event: KeyboardEvent) => {
+		if (event.key === 'Escape') handler()
```
micheal-parks, line 14: I believe we'll need to stop propagation here since there are other keybindings to this

src/lib/components/overlay/Popover.svelte
DTCurrie: I was hitting some weirdness but I can lower it. though IMO I think we may want to introduce some semantic z-layering classes to reduce having to go search the codebase for every z-x usage and determine how to layer. like z-popover, z-panel, z-label or something.

src/lib/plugins/Gizmos/useInputs.svelte.ts
DTCurrie: I tried both and from a user standpoint it feels less like "undo" and more like "go back" so the backspace felt natural, especially with space and enter for confirm and submit. but we can swap it out if we decide we want ctrl+z/ctrl+y to be the global undo/redo commands when we apply them more broadly.

src/lib/plugins/Gizmos/useInputs.svelte.ts
DTCurrie: do you know which? I didn't see any when searching. I can still stop propagation either way.

src/lib/components/overlay/Popover.svelte
micheal-parks: yep, all for this

PR #740, opened 2026-06-01
App 16360 add llm 3 d scene generation

sucrammal:
This is a prototype for an LLM-powered frame builder. It allows you to type in a natural language query to edit frame orientation, position, and parent of one or more components in the visualizer. 

Here's a short, imperfect demo that shows me moving an arm to a new position, as well as resetting it back to its origin. 

https://github.com/user-attachments/assets/ab022c10-1397-43ce-b22f-d83b0222576c

Here's another demo of using the frame builder alongside the other editor. Making edits in both ways creates a config diff together. 

https://github.com/user-attachments/assets/79582e7c-104d-4ed4-a08e-636fb7d47a8f

However, I'm having a little trouble with local position vs. world position. I had a peek at the worldPose/worldMatrix implementation, and heard from @mattmacf98 that there was an ongoing bug there. In this demo, an update from the frame builder, although confirmed, does not show a visual preview in its new position. Looking at the component edit tab shows a stale World Position that is only later updated. Not sure if this is because it is waiting for it to update from the network, or something else. 

https://github.com/user-attachments/assets/8c014024-29fd-4b03-8f56-a313025e4e35

## TODOs
- Investigate world position vs. local position discrepency bug. 
- Currently, the chat is single-shot (i.e. only accepts one prompt at a time, no conversation history). If it's a desirable feature, we can extend the chat functionality. 
- There is no place to place your own LLM provider key. When the visualizer is wired into the app, we can reuse the app/inference routes that are already there. But as suggested by @mattmacf98 , might be easier to have a manual entry in the settings widget for development. 
- General stress testing and robust support simultaneous editing (use frame builder, drag edit session, manual entry).


src/routes/api/scene-builder/+server.ts
```diff
@@ -0,0 +1,104 @@
...
+			z: z.number(),
+			th: z.number(),
+		})
+		.optional(),
+	parent: z.string().optional(),
+})
+
+const ResponseSchema = z.object({
+	updates: z.array(FrameDeltaSchema),
+	explanation: z.string(),
```
mattmacf98: nit: looks to work well but an explanation per delta might be worth checking out for very complex all in one commands to have the LLM plan steps more piece wise

server/routes/scene-builder.ts
```diff
@@ -0,0 +1,104 @@
+import { ChatAnthropic } from '@langchain/anthropic'
```
mattmacf98, line 1: svelte kit has built in api routes, you just need to put it in a +server.ts file

e.x. `src/routes/api/scene-builder/+server.ts`

server/routes/scene-builder.ts
```diff
@@ -0,0 +1,104 @@
...
+	// Full orientation replacement as ov_degrees { x, y, z, th }
+	orientation: z
+		.object({
+			x: z.number(),
+			y: z.number(),
+			z: z.number(),
+			th: z.number(),
+		})
+		.optional(),
+	parent: z.string().optional(),
```
mattmacf98, line 37: nit: it seems to work pretty well right now, but langraph + zod works really really well when you add `.describe()` to your fields

e.x. 

```
parent: z.string().describe("the component which is the parent of this frame, transforms will be applied in tree like hierarchy"),
```

src/routes/api/scene-builder/+server.ts
```diff
@@ -0,0 +1,104 @@
...
+			z: z.number().optional(),
+		})
+		.optional(),
+	// Full orientation replacement as ov_degrees { x, y, z, th }
+	orientation: z
+		.object({
+			x: z.number(),
+			y: z.number(),
+			z: z.number(),
+			th: z.number(),
```
mattmacf98: we can specify schema limit in zod e.x. `.min(0).max(1)` or maybe even the object level to enforce normalization

src/routes/api/scene-builder/+server.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const CORS_HEADERS = { 'Access-Control-Allow-Origin': '*' }
+
+const SYSTEM_PROMPT = `You are a robot spatial configuration assistant. The user wants to adjust frame positions and orientations of robot components. Return only the components that need to change and only the fields being changed (delta — do not repeat unchanged fields).
+
+Rules:
+- Only modify components listed in the context below. Each component has a "name" field — use that exact string as "componentName" in your response.
+- Return only components that actually need to change.
+- For translation, return only the changed axes (x, y, z are each optional).
+- For orientation, return the full { x, y, z, th } orientation vector: (x,y,z) is the unit axis, th is the rotation angle in degrees.
```
mattmacf98: orientation vector is a viam internal math thing so I would be surprised if LLMs know it, maybe we should swap to quaternions or euler angles

src/routes/+layout.svelte
```diff
@@ -72,10 +73,10 @@
 
 				{#snippet dashboard()}
 					<Machines bind:isOpen={isMachinesPageOpen} />
+					<SceneBuilder />
```
mattmacf98: why does this need to be in the dashboard snippet?

src/lib/components/App.svelte
```diff
@@ -101,6 +102,7 @@
 		() => partID,
 		() => localConfigProps
 	)
+	provideSceneBuilder()
```
mattmacf98: we will probably want to put this provide only in the LLMSceneBuilder plugin, look at how the draw service sets it up https://github.com/viamrobotics/visualization/blob/78579a275c19f11b864e4c1795bf9def91c5fbe1/src/lib/plugins/DrawService/DrawService.svelte#L2 as inspo

src/lib/components/overlay/SceneBuilder/frameDeltaAdapter.ts
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	return { errors, prepared }
+}
+
+/**
+ * Applies previously validated and prepared frame updates to the config.
+ * Call this on user confirmation after validateProposedFrameDeltas().
+ */
+export function applyPreparedUpdates(prepared: PreparedUpdate[], deps: Deps): void {
```
mattmacf98: I think this function should live entirely in the useSceneBuilder (that way you can just call the validateProposedFrameDeltas helper to get the preparedUpdate and then do the submit all in the context without having to pass back in this updateFrame deps back into here

mattmacf98 (commented):
This looks great! I left a couple of nits to think about when doing the final PR (I think this is non-invasive enough to just do one PR)

biggest thing is just to move this to a plugin see DrawService for an example on how to do that [visualization/src/lib/plugins/DrawService/DrawService.svelte](https://github.com/viamrobotics/visualization/blob/78579a275c19f11b864e4c1795bf9def91c5fbe1/src/lib/plugins/DrawService/DrawService.svelte#L2)

src/routes/api/scene-builder/+server.ts
sucrammal: Screenshot 2026-06-02 at 10 56 15 AM

How does this look? For the "all-in-one" prompts, I reorganized the output to display individual tables by component with a more involved explanation breaking down how it interpreted "unintuitive" parts of the prompt.

server/routes/scene-builder.ts
sucrammal: Sweet! Reafactored that in the dir you suggested – need to add endpoint into Bun server as well as local input of API key.

server/routes/scene-builder.ts
sucrammal: added describe() to fields that are not as explicitly defined.

src/routes/api/scene-builder/+server.ts
sucrammal: A few demos on rotation here. LLM interprets and outputs Euler angles. 

https://github.com/user-attachments/assets/8f268fc4-93d9-4c29-a0e9-2fea2137bcec

https://github.com/user-attachments/assets/c53b366a-2027-4bdd-8ead-759033456596

src/lib/components/overlay/SceneBuilder/frameDeltaAdapter.ts
sucrammal: Yep! I inlined the apply loop directly in confirm() here a9e10d6a83808779370e15bd26108e68536cbac6

src/lib/hooks/useFrames.svelte.ts
```diff
@@ -233,7 +240,26 @@ export const provideFrames = (partID: () => string) => {
...
+					// monitor mode. isDirty is $state and updates synchronously with
+					// updateFrame(), so it's always current here.
+					//
+					// !isDirty also handles the post-save case: standalone save() clears
+					// isDirty synchronously after the mutation resolves (at the same time
+					// hasPendingSave is set), so the baseline unlocks to the saved config
+					// as soon as the save completes. If the user edits again before the
+					// robot confirms (isDirty=true again), the baseline re-freezes,
+					// keeping the new preview correct.
+					if (!partConfig.isDirty) {
```
micheal-parks: Would it be possible to move any core fixes (not part of the LLM plugin) into its own PR? I'd like to carefully review that in isolation

src/lib/transform.ts
```diff
@@ -161,6 +161,30 @@ export const matrixToPose = (matrix: Matrix4, pose: Pose): Pose => {
  * `Frame.svelte` uses to blend live kinematics with user-staged edits;
  * `worldMatrix.ts` premultiplies the result by the parent's `WorldMatrix`.
  */
+export const poseToEulerDegrees = (pose: Partial<Pose>): { roll: number; pitch: number; yaw: number } => {
```
micheal-parks: Interesting, does the Typescript SDK not expose Viam's euler type signature?

src/lib/hooks/useFrames.svelte.ts
sucrammal: Yes – I can revert changes in this file and move it to a new PR. I believe all the world position / local position baseline matrix / world matrix issues were contained in here. This probably means the feature won't work too well, is that okay?

micheal-parks: Totally keep them here, I'd rather just merge them alone first after an independent review :)

sucrammal: Gotcha!

src/lib/transform.ts
sucrammal: Yes, the TS SDK includes the Euler type signature (https://ts.viam.dev/classes/appRobotApi.Orientation_EulerAngles.html). It's just that Orientation_eulerAngles is defined in radians. The callers of these functions are intentionally in degrees just because degrees are a more natural unit for the end user for frame editing. I could edit this function to use the SDK defined type signature – will just need to add conversions from deg <-> radians. what do you think?

src/routes/api/scene-builder/+server.ts
sucrammal: Since we switched to Euler angles for the LLM input and output. Clamped each field to [-180, 180].

src/lib/hooks/useFrames.svelte.ts
sucrammal: Here: #742 !

src/lib/hooks/useFrames.svelte.ts
```diff
@@ -89,8 +89,15 @@ export const provideFrames = (partID: () => string) => {
...
 		// never transition through DISCONNECTED).
+		//
+		// isDirty is included alongside didRecentlyEdit because didRecentlyEdit
+		// is set by a plain $effect (runs after DOM) and lags by one flush on
+		// the first edit from monitor mode. isDirty is $state and flips
+		// synchronously in updateFrame(), so configFrames becomes a tracked
+		// dependency of this derivation the moment the user makes any edit.
 		if (
 			didRecentlyEdit ||
+			partConfig.isDirty ||
```
mattmacf98: remove these changes, I will look into the flashing as part of https://viam.atlassian.net/browse/APP-16024

src/lib/components/overlay/settings/Settings.svelte
```diff
@@ -71,6 +72,7 @@
 			{ label: 'Vision', component: VisionSettings },
 			{ label: 'Widgets', component: WidgetSettings },
 			{ label: 'Debug', component: DebugSettings },
+			{ label: 'AI', component: AISettings },
```
mattmacf98: move AISettings into the plugin dir for AI and then inject it in via settingsTabs from the top level look at XRSettings as an example https://github.com/viamrobotics/visualization/blob/main/src/routes/%2Blayout.svelte#L70

mattmacf98 (commented):
very close! there are just a couple more extractions to be made to fully plugin-ize this

1. nothing should go in App.svelte your provideAnthropicKey, LLMSceneBuilder component etc... should just go in the +layout.svelte so they are used in local motion tools without baking them into the exported MotionTools component
2. pull the AISettings out and inject from the top layer

src/lib/components/App.svelte
```diff
@@ -175,6 +205,9 @@
 				<Settings {settingsTabs} />
 				<Logs />
 				<AddFrames />
+				{#if !localConfigProps || onInfer}
+					<LLMSceneBuilder onInfer={onInfer ?? standaloneInfer} />
```
mattmacf98: this should not go here, this App gets exported as the <MotionTools/> component, put it here https://github.com/viamrobotics/visualization/blob/main/src/routes/%2Blayout.svelte#L78-L81 like the other local plugins

src/lib/plugins/LLMSceneBuilder/SceneBuilder.svelte
```diff
@@ -0,0 +1,145 @@
...
+		<DashboardButton
+			active={isOpen}
+			icon="robot-outline"
+			description="Frame Builder"
+			onclick={() => (isOpen = !isOpen)}
+		/>
+	</fieldset>
+</Portal>
+
+<FloatingPanel
```
mattmacf98: we can actually put this in `<Portal id="dom">` I just found out (instead of having to put in the dashboard snippet

mattmacf98 (commented):
LGTM 3 things before merge

1. run npx @changesets/cli to create a .changeset minor bump
2. remove the changes in [‎src/lib/hooks/useFrames.svelte.ts‎](https://github.com/viamrobotics/visualization/pull/740/changes#diff-442a6691b3a0fa760ea456f1c1f003259e7b25db662d562bf37619e601a2c5bf)
3. wrap the floatingPanel in `<Portal id="dom">` instead of putting in the dashboard snippet

mattmacf98 (approved):

mattmacf98 (approved):

PR #750, opened 2026-06-05
Extract logs into a plugin

micheal-parks:
### Overview

This PR does a partial migration of logs into a plugin. Logging calls still happen in core, because extracting them out would have been a much larger and longer change. This accomplishes most of our goals: the logging UI / behavior is ripped out of core and logging no-ops if the plugin isn't installed. We can do a migration later, and it might be best to do it later when machine connection is pulled out.


src/lib/plugins/Logs/useLogs.svelte.ts
```diff
@@ -83,9 +84,19 @@ export const provideLogs = () => {
...
 			})
 		},
-	})
+	}
+
+	return context
 }
 
-export const useLogs = () => {
-	return getContext<Context>(key)
```
mattmacf98, line 90: why can't we still use the svelte context and just make useLogs do 


```
export const useLogs = (): Context => {
    const context = getContext<Context>(key);
	// return a no-op context if the plugin isn't installed
	return (
		context ?? {
			current: [],
			errors: [],
			warnings: [],
			add: () => undefined,
		}
	)
}
```

mattmacf98 (approved):

DTCurrie (approved):

src/lib/plugins/Logs/useLogs.svelte.ts
micheal-parks: Then `<Logs>` would have to be a provider component. I'd like to avoid that considering in the future the full extraction of logs will make it so that it doesn't have to be a provider. I figure then we can switch back to a context

PR #751, opened 2026-06-05
Force directed graph labels

micheal-parks:
### Overview

Updates labels to use d3's force directed graph plugin to prevent overlap. For example, here's a complicated scene:

Screenshot 2026-06-05 at 13 21 09

This is a hard problem - Claude did a lot of algorithmic heavy lifting - and this solution is nowhere near perfect, it's just a lot better than the current label behavior, which renders them unusable pretty quickly as scene members grow.


src/lib/components/Entities/Entities.svelte
```diff
@@ -60,56 +61,48 @@
...
-	<GLTF {entity}>
-		<Label text={entity.get(traits.Name)} />
-	</GLTF>
+	<GLTF {entity} />
 {/each}
 
 <Arrows />
+
+{#if enableLabels}
+	<Labels />
```
mattmacf98, line 107: Nice!

mattmacf98 (approved):
The end product LGTM! The math is a lot but probably ok (wonder if force directed UI layouts might have a used case for general public and we could lift into @threlte/extras or something asp)

src/lib/components/Entities/Entities.svelte
micheal-parks: Yep! Labels are much less invasive when disabled now :)

PR #773, opened 2026-06-12
Instanced rendering of all boxes

micheal-parks:
### Overview

Migrates rendering of all box geometries to one `InstancedMesh2`. 

This PR should serve as the template for how we handle instanced rendering for other geometries in the future.


src/lib/components/Entities/hooks/useEntityEvents.svelte.ts
```diff
@@ -138,6 +148,39 @@ export const useEntityEvents = (entity: () => Entity | undefined) => {
...
+ * `onpointerleave` is intentionally left active. The effect tears down a stale
+ * Hovered/InstancedMatrix for an entity that turns invisible while hovered,
+ * since the guarded handlers can no longer fire to clean it up.
+ */
+export const useEntityEvents = (entity: () => Entity | undefined) => {
+	const cursor = useCursor()
+	const invisible = useTrait(entity, traits.InheritedInvisible)
+	const events = createEntityEvents(entity, cursor)
+
+	const whenVisible =
```
mattmacf98, line 177: nice!

src/lib/plugins/Focus/FocusBox.svelte
```diff
@@ -35,6 +37,8 @@
 
 	const box = new Box3()
 	const vec = new Vector3()
+	const unitBox = new Box3(new Vector3(-0.5, -0.5, -0.5), new Vector3(0.5, 0.5, 0.5))
```
mattmacf98, line 40: this is fine but kindof hidden that this get squashed down using `0.001` by `composeBoxMatrix` could we not just do the scale here?

src/lib/components/Entities/Boxes.svelte
```diff
@@ -0,0 +1,292 @@
...
+	 * instance matrix.
+	 */
+	const instancedBoxes = new InstancedMesh2(
+		unitBox,
+		new MeshToonMaterial({ transparent: true, depthWrite: false }),
+		{ renderer }
+	)
+	instancedBoxes.sortObjects = true
+	instancedBoxes.customSort = createRadixSort(instancedBoxes)
+	instancedBoxes.frustumCulled = false
```
mattmacf98, line 65: why no frustum culling?

src/lib/components/Entities/Boxes.svelte
```diff
@@ -0,0 +1,292 @@
...
+	)
+
+	const position = new Vector3()
+	const quaternion = new Quaternion()
+	const scale = new Vector3()
+	const matrix = new Matrix4()
+	const colorUtil = new Color()
+
+	/** Same resolution order as `Frame.svelte`. */
+	const resolveColor = (entity: Entity): Color => {
```
mattmacf98, line 113: nit: this should probably go into a general color helper asp

src/lib/components/Entities/Boxes.svelte
```diff
@@ -0,0 +1,292 @@
...
+		 * pointer moves, so drop its hover state here.
+		 */
+		if (!visible && entity.has(traits.Hovered)) {
+			entity.remove(traits.Hovered)
+		}
+	}
+
+	/** Caller composes the instance transform into `matrix` first. */
+	const addInstance = (entity: Entity) => {
+		matrix.decompose(position, quaternion, scale)
```
mattmacf98: why decompose? can we not add an instance with just the transform marix?

src/lib/components/Entities/Boxes.svelte
```diff
@@ -0,0 +1,292 @@
...
+		dirty.add(entity)
+		schedule()
+	}
+
+	const enqueueTransform = enqueue(dirtyTransform)
+	const enqueueAppearance = enqueue(dirtyAppearance)
+
+	$effect(() => {
+		for (const entity of world.query(traits.Box)) {
+			dirtyTransform.add(entity)
```
mattmacf98, line 245: why split these up if they just get enqueued and processed together anyway?

src/lib/components/Entities/Boxes.svelte
micheal-parks: InstancedMesh2 has per-instance frustum culling :)

src/lib/components/Entities/Boxes.svelte
micheal-parks: They're only processed together here, in the one-time initial sync. At runtime the two sets diverge by design: WorldMatrix fires changed() for every box every kinematics tick, so a moving robot enqueues transform work only.

src/lib/components/Entities/Boxes.svelte
micheal-parks: going to be refactoring colors in a bit so that this resolution isn't even necessary.

src/lib/components/Entities/Boxes.svelte
micheal-parks: simplified

src/lib/plugins/Focus/FocusBox.svelte
micheal-parks: made this more explicit

mattmacf98 (approved):

PR #782, opened 2026-06-16
SCOPE: Motion plan replay plugin

sucrammal:
## Changes
- Adds `useMotionPlanReplayer.svelte.ts` context managing plan list state (add/remove/select/step) and driving playback through `reconcileSnapshotEntities`
- Adds `parse-plan.ts` (zod-validated extraction of `frame_system/trajectory `from concatenated plan JSON chunks), `build-frame-descriptors.ts` (resolves static vs. revolute-joint frames, handles quaternion/euler/OV orientations, model-frame reparenting into per-step transforms), and `plan-to-snapshots.ts` (renders trajectory steps into `Snapshot[]`)
- Adds `MotionPlanReplayerUI.svelte` (floating panel – upload, list, select, remove) and `MotionPlanReplayerScrubber.svelte` (play/pause/step/seek transport), wired into `+layout.svelte` and exported from `plugins/index.ts`
- Adds `PartOfPlan` relation (`relations.ts`) tying every entity spawned for a plan, including nested model assets, to a single plan-root entity, so switching, clearing, or removing a plan cascades cleanup via koota's autoDestroy instead of manual entity bookkeeping
- Tests: Adds specs for parse-plan, build-frame-descriptors, plan-to-snapshots, and the PartOfPlan relation

## Links:
- [APP-9316](https://viam.atlassian.net/jira/software/c/projects/APP/boards/81?assignee=712020%3A809c367c-3308-4462-b085-acb400cafaf2&selectedIssue=APP-16206)
- [APP-9317](https://viam.atlassian.net/browse/APP-9317)

## Testing:
- Open the visualizer, click the "Motion Plan Replayer" dashboard button (or drag-and-drop a plan JSON file) to open the panel
- Feel free to use some [sample plans here](https://drive.google.com/drive/folders/1KPks3TaWhWAhIVu41bj_MmGMI4dci9Kh?usp=sharing). 
- Upload a valid motion plan JSON (frame_system + trajectory) → it appears in the list and the scrubber appears at the bottom
- Press play → frames animate through the trajectory; step forward/back and drag the scrubber → poses update correctly at each step
- With one plan active, select a second plan → the first plan's entities (including any model geometry) are fully removed before the second renders
- Remove the active plan, or click Exit on the scrubber → no leftover entities remain in the scene
- Drop/upload a malformed or non-plan JSON file → a toast error appears and nothing is spawned
- Upload a file with a name that's already loaded → the "already loaded" warning toast appears

**POC demo:**

https://github.com/user-attachments/assets/43b5eae8-8f6b-46b7-bbca-9f9b3984fa79

**Salad1:**

https://github.com/user-attachments/assets/1684b5ad-bc98-4b5e-8d16-b5dd72799dbb

**Vino:**

https://github.com/user-attachments/assets/fc083a2b-d1d0-48b0-957b-b9c843c61711

**Cappucina**

https://github.com/user-attachments/assets/9e425df8-8341-4605-9376-553b63a7a6c3


src/lib/components/FileDrop/FileDrop.svelte
```diff
@@ -71,6 +71,12 @@
 					)
 					break
 				}
+				case 'plan': {
+					window.dispatchEvent(
```
mattmacf98: why using custom events to send the result?

src/lib/components/FileDrop/useFileDrop.svelte.ts
```diff
@@ -25,6 +27,9 @@ const createFileDropper = (extension: string, prefix: string | undefined) => {
 		case Extensions.PLY: {
 			return plyDropper
 		}
+		case Extensions.JSON: {
```
mattmacf98: this is a little strict, we may have different jsons we want to drop in here, could we say that plan jsons have a required prefix e.x. `motion_plan_` to be droppable and do it the same way snapshots work

src/app.css
```diff
@@ -67,3 +67,8 @@ select {
 .motion-tools-table tbody th {
 	@apply border-light font-roboto-mono text-default h-[40px] gap-2 border px-1.5 text-center text-xs font-normal;
 }
+
+/* Push toasts above the motion plan scrubber bar (fixed bottom-4, ~44px tall). */
+body.has-scrubber [aria-label='Toasts'] {
```
mattmacf98, line 72: todo: for final impl we probably should look into how to do this with component placements instead of the app css

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -0,0 +1,230 @@
...
+	let currentStep = $state(0)
+	let entityMap = $state.raw(new Map<string, SnapshotEntity>())
+
+	const totalSteps = $derived(
+		activePlanIndex === null ? 0 : (plans[activePlanIndex]?.stepCount ?? 0)
+	)
+
+	const clearActivePlan = () => {
+		for (const entry of entityMap.values()) {
+			if (world.has(entry.entity)) hierarchy.destroyEntityTree(world, entry.entity)
```
mattmacf98: for the final setup, lets look into making a new PlanEntity, which we can then give relationships to all the world entities that are part of it (see the selection plugin https://github.com/viamrobotics/visualization/blob/main/src/lib/plugins/Selection/traits.ts for inspo)

this way we can make this flow more koota native and the relationship has an auto destroy hook e.x. https://github.com/viamrobotics/visualization/blob/main/src/lib/plugins/Selection/relations.ts#L9 so we can make it so all world entities auto clean up when a plan entity is removed

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -0,0 +1,230 @@
...
+			if (!entry.entity.isAlive()) continue
+			preserved.set(entry.entity, {
+				opacity: entry.entity.get(traits.Opacity) ?? 1,
+				invisible: entry.entity.has(traits.Invisible),
+			})
+		}
+
+		const result = reconcileSnapshotEntities(world, snap, entityMap)
+
+		for (const spawned of result.spawned) {
```
mattmacf98, line 92: what is this loop trying to do? is it manually trying to apply relationships added in the reconcile step?

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -0,0 +1,230 @@
...
+			if (uuid) relationships.flush(uuid)
+		}
+		entityMap = result.current
+
+		// Spawned entities (first appearance): apply plan defaults.
+		// Apply color only to geometry entities; apply opacity to all plan entities
+		// so that reference frame entities also start at the plan opacity (not 1.0).
+		for (const entry of result.spawned) {
+			if (!entry.entity.isAlive()) continue
+			if (!entry.entity.has(traits.ReferenceFrame)) {
```
mattmacf98: assuming these wont have a reference frame might work for motion plans but could get fragile if we do ever have nested entities in motion plans, it would be good when we have the Plan entity and a relationship for `PartOfPlan` we can just query for the entities that are parts of a plan and do this logic

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -0,0 +1,230 @@
...
+
+		// Capture user-modified opacity and visibility before reconcile.
+		// updateTransform → updateMetadata unconditionally resets Opacity to DEFAULT (1.0)
+		// and removes Invisible — both of which erase user changes made via the legend.
+		// We restore the captured state on updated entities after reconcile.
+		const preserved = new Map<Entity, EntityState>()
+		for (const entry of entityMap.values()) {
+			if (!entry.entity.isAlive()) continue
+			preserved.set(entry.entity, {
+				opacity: entry.entity.get(traits.Opacity) ?? 1,
```
mattmacf98: the maintaining of opacity and visibility is cool! but for the first stage lets start simple and not maintain and merge this (should simplify the initial code a lot) 

I think this merging of snapshot and user toggled data can be addressed in a future improvement using update snapshots (something we need to discuss for v2 of snapshots)

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -0,0 +1,230 @@
...
+type EntityState = { opacity: number; invisible: boolean }
+
+const KEY = Symbol('motion-plan-replayer')
+
+export const provideMotionPlanReplayer = (initialPlans?: PlanEntry[]) => {
+	const world = useWorld()
+	const relationships = useRelationships()
+
+	// Proto objects stored here — never inside $state to avoid Svelte 5 deep proxy
+	const snapshotStore = new Map<number, Snapshot[]>()
```
mattmacf98, line 51: idt there is a huge use case for storing multiple plans at once (unless the plan --> snapshot conversion is really slow) lets start with just one in memory at a time so just and array of SnapShot[]

src/lib/plugins/MotionPlanReplayer/plan-to-snapshots.ts
```diff
@@ -0,0 +1,86 @@
...
+				referenceFrame: descriptor.parent,
+				pose: descriptor.localPose,
+			}),
+			physicalObject: descriptor.geometry ?? undefined,
+			uuid: descriptor.uuid,
+		})
+	}
+
+	const angleRad = stepInputs[descriptor.componentName]?.[descriptor.jointIndex] ?? 0
+	return new Transform({
```
mattmacf98, line 47: what are non static ones? I thought these were the ones we wanted to drop?

src/lib/plugins/MotionPlanReplayer/plan-to-snapshots.ts
```diff
@@ -0,0 +1,86 @@
...
+	descriptors: FrameDescriptor[],
+	trajectory: Array<Record<string, number[]>>
+): Snapshot[] =>
+	trajectory.map((stepInputs) => {
+		const transforms = descriptors.map((d) => descriptorToTransform(d, stepInputs))
+		return new Snapshot({ transforms, uuid: planUuid() })
+	})
+
+export const planJsonToSnapshots = (content: string): Snapshot[] => {
+	const plan = parsePlan(content)
```
mattmacf98: nit: I would have this take in the already parsed plan instead of a raw string

src/lib/plugins/MotionPlanReplayer/plan-dropper.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Adapted detection from POC: bov-debug-bad-plans:src/lib/loaders/plan-request-loader.ts
+export const planDropper: FileDropper = async ({ name, content }) => {
+	const label = truncate(name)
+
+	if (typeof content !== 'string') {
+		return { success: false, error: new FileDropperError(`"${label}" failed to load.`) }
+	}
+
+	if (!content.includes('"frame_system"')) {
```
mattmacf98: probably good to do this step in parsePlan

src/lib/plugins/MotionPlanReplayer/plan-dropper.ts
```diff
@@ -0,0 +1,31 @@
...
+		return {
+			success: false,
+			error: new FileDropperError(`"${label}" is not a valid motion plan JSON.`),
+		}
+	}
+
+	try {
+		const snapshots = planJsonToSnapshots(content)
```
mattmacf98: we should call parsePlan here and pass the validated pieces into planJsonToSnapshots

src/lib/plugins/MotionPlanReplayer/parse-plan.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const parsePlan = (content: string): ParsedPlan => {
+	let frames: Record<string, RawFrame> = {}
+	let parents: Record<string, string> = {}
+	let trajectory: Array<Record<string, number[]>> = []
+	let goals: unknown[] = []
+	let foundFrameSystem = false
+
+	let index = 0
+	for (let i = 0; i < 8; i++) {
```
mattmacf98, line 80: are the motion plan jsons not valid json? If they are valid, we should just be able to do a big JSON.parse(content) and then use zod to confirm it fits our expected format

```
const UserSchema = z.object({
  name: z.string(),
  age: z.number(),
})

const user = UserSchema.parse(JSON.parse(json))
```

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerUI.svelte
```diff
@@ -0,0 +1,94 @@
...
+		const handle = (e: Event) => {
+			const { name, content, snapshots } = (e as CustomEvent<PlanFileDropSuccess>).detail
+			if (ctx.plans.some((p) => p.name === name)) {
+				toast({ message: `"${truncate(name, 24)}" already loaded.`, variant: ToastVariant.Warning })
+				return
+			}
+			ctx.addPlan(name, content, snapshots)
+			isOpen = true
+		}
+		window.addEventListener('viam:plan-loaded', handle)
```
mattmacf98: Ideally we do this in a different way when we actually build out (maybe plugins need to be able to reach into drag and drop and create custom drag and drop fucntionality?)

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerScrubber.svelte
```diff
@@ -0,0 +1,166 @@
...
+		if (intervalId !== undefined) {
+			clearInterval(intervalId)
+			intervalId = undefined
+		}
+	}
+
+	const play = () => {
+		if (isPlaying || ctx.totalSteps <= 0) return
+		isPlaying = true
+		intervalId = setInterval(() => {
```
mattmacf98: I think it would be nice if isPlaying if we set up the interval as an effect on isPlaying so it does something like this


```
const play = () => { isPlaying = true}

$effect(() => {
   if (isPlaying) {
      intervalId = setInterval(() => {...}, STEP_INTERVAL_MS)
   }

   return () => {
       clearInterval(intervalId)
   }
})
```

this way the interval automatically starts when we are playing and clears itself and stops when we stop playing

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayer.svelte
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	const { plans }: Props = $props()
+
+	// `plans` is intentionally read once at setup time — initial entries only.
+	// The context manages its own reactive plan list after that.
+	provideMotionPlanReplayer(untrack(() => plans))
+</script>
+
+{#if plans === undefined}
```
mattmacf98: it would be cool if drag and drop just updated the plans array and then we could use the ui for both cases

src/lib/plugins/MotionPlanReplayer/interpolate-trajectory.ts
```diff
@@ -0,0 +1,45 @@
+const INTERP_STEP_RAD = (5 * Math.PI) / 180 // 5° per sub-step
+
+/**
+ * Densify a trajectory by linearly interpolating between keyframes.
+ *
+ * Sub-step count between each pair is driven by the joint with the largest
+ * angular change, so all joints move at uniform velocity relative to each other.
+ */
+export const interpolateTrajectory = (
```
mattmacf98: I don't really think we can do this reliably since we don't actually know what path the arm takes between frames (i.e. might not be a linear interpolation)

mattmacf98 (commented):
Great POC! looks very nice!

my main things for when we do the full impl are
1. we should try to be more koota native and create a plan entity and relationships for the plan <--> plan entities
2. I think we can use zod to parse the plans a little easier
3. we should strip back some of the (very cool) functionality you added in P0 (i.e. no linear interp between steps, no preserve opacity and visibility (yet), no multiple plans loaded at a time)

src/lib/components/FileDrop/useFileDrop.svelte.ts
sucrammal: Resolved in 7930ad6042244549fee6e0544996a99c46f59fc9. Moved all file uploading into plugin dir. Currently does not accept plans or any JSON if dragged and dropped into wider viz canvas.

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerUI.svelte
sucrammal: Simplified in 7930ad6042244549fee6e0544996a99c46f59fc9!

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Reverted for now!

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerScrubber.svelte
sucrammal: Gotcha. That makes sense! Keep isPlaying as the only moving piece, and isEffect off of that for a singular interaction with the timer.

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Hmmm I'm not entirely sure about the use case either, but even from taking a look at some of the motion plans from Vino, many different motion plans make up a full flow. For example, dropping the arm down to cup level --> approaching the cup. I could imagine that if you wanted to debug a coupled motion like that, I would want to switch between 1-3 motion plans at a time?

src/lib/plugins/MotionPlanReplayer/parse-plan.ts
sucrammal: Yes with a small caveat: motion plan files are concatenation of 2 valid JSON objects: the request (frame_system plus start and end poses) plus the motion plan (plan + trajectory). So there is still a little bit of non-zod work to dissect the file into these objects.

src/lib/plugins/MotionPlanReplayer/plan-dropper.ts
sucrammal: Done! Corrected the full caller --> parse flow.

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayer.svelte
sucrammal: Designed a more explicit hook here (f3f70cb8763be54040eac50a6401d5d7b20d4d37) which calls the same addPlan hook from app vs. standalone.

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Yea its manually rewiring up parent-child relationships that `reconcileSnapshotEntities` computed but didn't apply yet. The caller (`result.spawned[i].relationships` actually applies these relationships for the ECS.

src/lib/plugins/MotionPlanReplayer/plan-to-snapshots.ts
sucrammal: TLDR static descriptors are frames with no non-static, jointed parent. Think obstacle. The implicit else block is for a custom frame type called `jointed_links` which are frame descriptors with a static link with a rotational, non-static link as a parent. This complicated parsing is all in an effort to not implement full world-space FK.

Here's an example simplified frame: 
```
  left-arm:base       (static link, at arm origin)
  --> left-arm:waist    (rotational joint, Z-axis)  
  --> --> left-arm:base_top  (static link, translation X=50)  
```
By default, the parents would be:
```
"left-arm:waist"    → "left-arm:base"
"left-arm:base_top" → "left-arm:waist"
```
But we don't want a link to be parented towards a joint with no geometry like we discussed (this is the "skipping" of non-static links). They aren't in the ECS at all. However, we use the trajectory of the rotational joint (e.g. how many degrees does it move in step 0?) and apply it to the child link (e.g. waist and base_top in our example) to compute its new pose. I collapsed these 2 objects into a custom `jointed_link` descriptor, which is what this else code block is handling. 

This was one of the way I thought of how to keep the ECS simple while maintaining the dynamics of an arm. We should def take a look at the underlying algo that builds frames into snapshots more closely...

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Tried mirroring the pattern from Selection plugin here: 5631b12b4f90b280610209f18274399a2ea8f6f2. This makes the destroy plan path logically clearer! Is this up to your spec?

mrloureed (changes requested):
added comment to the ticket for design feedback

PR #800, opened 2026-06-30
Improve plugins infrastructure

micheal-parks:
### Overview

This PR introduces 3 new portaling components for plugins: `DashboardPortal`, `DetailsPortal` and `SettingsPortal`. The last one is the most configurable and allows creation of settings tabs. The first two are pretty barebones and only allow what we currently do with injecting into the dashboard and details panels, but in subsequent PRs I'll improve them so that we can order dashboard items and allow injection into specific areas of the details panel.

Additionally, this PR updates floating panels so that they are automatically wrapped in DOM portals, removing the need for plugins to do this.


src/lib/components/overlay/Portals/SettingsPortal.svelte
```diff
@@ -0,0 +1,34 @@
...
+		children: Snippet
+	}
+
+	let { label, children }: Props = $props()
+
+	const tabs = useSettingsTabs()
+
+	$effect(() => {
+		// eslint-disable-next-line @typescript-eslint/no-unused-expressions
+		label
```
mattmacf98: nit: would pulling `const tab = { label, children }` up here turn on reactivity for these so we don't need to do the eslint ignore

src/lib/components/overlay/settings/Tabs.svelte
```diff
@@ -1,16 +1,22 @@
...
-			component: Component
-		}[]
+		items: (
+			| {
+					label: string
+					component: Component
+			  }
+			| {
+					label: string
+					children: Snippet
```
mattmacf98, line 17: why does this prop need to render children now? can we not just have the useSettingsTab require a Component or move everyone over to use snippets?

micheal-parks: The problem is that components make sense when making tabs internally, and snippets make sense ergonomically when making tabs via plugins. Small complexity cost internally.

src/lib/components/overlay/Portals/SettingsPortal.svelte
micheal-parks: good point, fixed!

mattmacf98 (approved):

DTCurrie (approved):

PR #802, opened 2026-06-30
Instance render capsules

micheal-parks:
### Overview

After boxes and AxesHelpers, capsules are often the next most rendered geometry object. This PR switches them over to instanced rendering using the exact same technique that we established for boxes. After this PR, the only remaining instance-able geometric primitive left as a todo are spheres.


src/lib/components/Entities/Boxes.svelte
```diff
@@ -121,8 +119,7 @@ on each hit, which `useInstancedEntityEvents` maps back to the entity.
 			return colorUtil.setRGB(color.r, color.g, color.b)
 		}
 
-		const subtype = resourceByName.current[entity.get(traits.Name) ?? '']?.subtype
-		return subtypeToColor(subtype) ?? colorUtil.set(colors.default)
+		return colorUtil.set('gray')
```
mattmacf98: nit: why not `colors.defualt`

mattmacf98 (approved):

src/lib/components/Entities/Boxes.svelte
micheal-parks: good point!

PR #805, opened 2026-07-01
App 17034 add geo editing for llm scene builder

sucrammal:


server/routes/scene-builder.ts
```diff
@@ -34,6 +47,19 @@ const FrameDeltaSchema = z.object({
 			yaw: z.number().min(-180).max(180).optional(),
 		})
 		.optional(),
+	geometry: z
+		.object({
+			type: z.enum(['none', 'box', 'sphere', 'capsule']).optional(),
```
mattmacf98, line 52: geometry itself should be optional but I think a given delta should have the type enum defined

src/lib/plugins/LLMSceneBuilder/frameDeltaAdapter.ts
```diff
@@ -97,6 +114,74 @@ function mergeOrientation(
...
 
+function mergeGeometry(
+	a?: FrameDelta['geometry'],
+	b?: FrameDelta['geometry']
+): FrameDelta['geometry'] {
+	return a || b
+		? {
+				type: b?.type ?? a?.type,
+				x: b?.x ?? a?.x,
+				y: b?.y ?? a?.y,
```
mattmacf98: I think we need to merge based on different types right now this looks like if we went from a box with x:10 y:10 z:10 to a sphere with r:10 we would merged in the xyz

sucrammal: Thanks for the catch that was careless. Copied the translation and orientation merge patterns too closely. Fixing now: dd003c6c85ec857c56fe26ce80d4e6bb6a6ac747

server/routes/scene-builder.ts
sucrammal: The type being optional here is intentional on my part. If we are just trying to say, update a box geometries x field, the model is currently just going to send a resize-only delta (e.g. x: 250) without repeating the shape. The describe() field helps to describe this already. If you think we should always re-emit the type, I'll add it in right now.

mattmacf98: oh yes that makes sense ignore me

mattmacf98 (approved):
LGTM!

PR #801, opened 2026-06-30
App 16362 update snapshot api to v2 style inputs returns

sucrammal:
## Changes
- draw/snapshot.go: Replaces positional-arg signatures on DrawGeometry, DrawFrame, DrawFrameSystemGeometries, DrawArrows, DrawLine, DrawModel, DrawPoints with DrawXxxOptions structs returning ([]byte, error) / ([][]byte, error); adds DrawNurbs, DrawPointCloud, DrawGeometriesInFrame, DrawFrames, DrawWorldState; adds snapshotDrawableOpts helper to consolidate UUID/parent/pose/attrs wiring across all methods; appends all 12 options structs to the same file following existing draw/ conventions

## Tests: 
- Updates all snapshot_test.go call sites to DrawXxxOptions structs; updates docs/examples/snapshot/main.go fixture

## Links
- [APP-16362](https://viam.atlassian.net/jira/software/c/projects/APP/boards/81?assignee=712020%3A809c367c-3308-4462-b085-acb400cafaf2&selectedIssue=APP-16362)

## Docs: 
- Adds draw.Snapshot: static scene builder section to migration/v1-to-v2.mdx with quick-reference table and before/after examples; fixes stale v1 call in guides/embedding.mdx and draw/doc.go; regenerates docs/src/content/docs/api/draw.md via pnpm --dir docs gen:api
- static/visualization_snapshot.json: Regenerated from updated example

[APP-16362]: https://viam.atlassian.net/browse/APP-16362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #808, opened 2026-07-02
scene builder UI improvements

sucrammal:
These are the minimum fixes that Steve wanted so that we can put this feature back out and get feedback ASAP. Will consider doing accordion/selection style tables to make the information less crammed for multi-component diffs. 
- Explicit old vs. new diff headers, using prime colors used in the AI config assistant diffs. 
- "confirm" button --> "apply" for something more active. 
- Tooltip for cut off component names and diff explanations
- Added min width and min height for the floating panel to prevent over-shrinking.
- MISSING HERE: Clearer error state for instructions that try to add components. Since this is a deeper LLM config change, this was done in a separate PR #801 

Demo of a large instruction: 

https://github.com/user-attachments/assets/54e7e575-8e2a-421d-bf67-5ce3a8e3e0ee

### DIFF
--- a/.changeset/hungry-breads-end.md
+++ b/.changeset/hungry-breads-end.md
@@ -1,0 +1,5 @@
+---
+'@viamrobotics/motion-tools': patch
+---
+
+UI/UX LLM scene builder improvements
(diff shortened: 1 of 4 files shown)


mattmacf98 (approved):
LGTM

PR #810, opened 2026-07-02
Allow mesh vertex colors

micheal-parks:
### Overview

`ply` files can have per vertex colors specified. Right now we ignore this, limiting the expressiveness of any meshes that contain meaningful vertex data. This PR adds per vertex color support:

Screenshot 2026-07-02 at 16 28 10


mattmacf98 (approved):
pretty!

DTCurrie (approved):

PR #817, opened 2026-07-06
Fix pointcloud object positions

micheal-parks:
### Overview

At some point in the instanced rendering PR stack, I missed pointcloud objects, causing them to not be displayed in the correct location currently. This PR simply brings them up to speed with exactly how geometries are assembled in `useGeometries`, fixing any transform issues.

Screenshot 2026-07-06 at 13 07 47


mattmacf98 (approved):
LGTM

PR #820, opened 2026-07-06
Cleanup geom typing

sucrammal:
## TLDR
FrameDelta['geometry'].type duplicated the four geometry-shape literals by hand instead of referencing FrameGeometry, and resolveGeometry/createGeometryFromFrame matched on shape type via if chains that silently fell through to a runtime error string or undefined on an unhandled case. This makes FrameGeometry (from frame.ts) the single source of truth for the delta's type field and converts both functions to exhaustive switch statements, so adding a new geometry shape (future additions) now fails to compile at both call sites until handled.

## Changes
-  Types FrameDelta['geometry'].type as FrameGeometry instead of a duplicated literal union
-  Converts resolveGeometry (frameDeltaAdapter.ts) and createGeometryFromFrame (geometry.ts) from if chains to exhaustive switch statements with never-check defaults

## Testing
- Run pnpm check and pnpm lint: both clean.
- Run pnpm vitest run: all 282 tests pass, including frameDeltaAdapter.spec.ts.

### DIFF
--- a/.changeset/fluffy-readers-rush.md
+++ b/.changeset/fluffy-readers-rush.md
@@ -1,0 +1,5 @@
+---
+'@viamrobotics/motion-tools': patch
+---
+
+Tighten geom typing (for LLM scene builder)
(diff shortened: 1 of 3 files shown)


mattmacf98 (approved):
lgtm

PR #822, opened 2026-07-07
App 15451 Adds a fullscreen toggle to the visualizer when it is embedded in a host app,

mrloureed:
Adds a fullscreen toggle to the visualizer when it is embedded in a host app, so users
working in the Viam app's 3D scene tab can expand the scene to fill the browser window.

### Hoverd states
image
image
image

### Frontend

- **`FullscreenButton.svelte`** (new): a button in the top right of the scene that toggles
  fullscreen, styled to match the other overlay buttons (white chip, `lucide` icons, prime
  `Tooltip`). The icon swaps between `Maximize2` and `Minimize2`, `Escape` exits, and the
  host page's scroll is locked while fullscreen so only the visualizer receives it.
- **`App.svelte`**: derives `isEmbedded` from `localConfigProps` and renders the button only
  when embedded. When fullscreen, the root container switches from `relative` to
  `fixed inset-0 z-max`. The root background is now an unconditional `bg-white`
  (previously `dark:bg-white` only).
- **`App.svelte`**: selected-entity details cards get a 38px base offset when embedded
  (button height plus gap) so they open below the fullscreen button instead of under it.
- **`FullscreenButton.svelte.spec.ts`** (new): covers the click toggle, scroll lock and
  restore, and Escape-to-exit.

### Why?

**Why gate on `localConfigProps` instead of `environment.isStandalone`?**

`environment.current.isStandalone` is written in an `$effect`, so it is stale during the
initial render. Deriving `isEmbedded` from the prop directly is synchronous and uses the
same source of truth.

**Why CSS `fixed` instead of the browser Fullscreen API?**

It matches how our docs examples do it, keeps the browser chrome available, and avoids the
permission and lifecycle quirks of `requestFullscreen`. `z-max` (1000) sits above all app
chrome.

**Why make `bg-white` unconditional?**

The WebGL canvas is transparent and previously relied on the host page behind it being
white; dark mode already forced `dark:bg-white`. In fullscreen the overlay sits above the
app's nav and tabs, which bled through the transparent canvas. Painting the root white
always is simpler and removes that entire class of bug.

### Testing

Ran `pnpm test:unit-run` (three new tests in `FullscreenButton.svelte.spec.ts`),
`pnpm check`, and `pnpm lint`.

Manual verification, embedded in the app:

1. Build and link into a locally running app stack:
   `pnpm package` here, then in the app repo
   `make link-local PKG=@viamrobotics/motion-tools DIR=../visualization` and hard-reload.
2. Open a machine and go to the **3D SCENE** tab. A fullscreen button appears in the top
   right of the scene.
3. Click it: the scene covers the full window with an opaque background (no app nav
   bleeding through) and the icon changes to minimize.
4. Press `Escape` (or click again): the scene returns to the tab layout.
5. Select a component in the scene or world tree: the details card opens below the
   fullscreen button, not under it.

Manual verification, standalone: `make up` in this repo. No fullscreen button renders and
the details card position is unchanged.


src/lib/components/App.svelte
```diff
@@ -150,14 +167,18 @@
...
 					/>
 				{/each}
 
 				{#if environment.current.isStandalone}
 					<LiveUpdatesBanner />
 				{/if}
 
+				{#if isEmbedded}
+					<FullscreenButton bind:fullscreen />
+				{/if}
```
micheal-parks: We should probably make this a plugin instead. Check out the routes/layout.svelte file and look at the other plugins like `<MeasureTool />` to see how this is done

mrloureed: yea see Devin's comment

src/lib/components/App.svelte
```diff
@@ -92,6 +93,10 @@
 
 	const settings = provideSettings()
 	const environment = provideEnvironment()
+	const fullscreen = provideFullscreen()
+
+	// Clear the Fullscreen plugin's button (30px + 8px gap) so details cards don't cover it.
+	const detailsBaseOffset = $derived(fullscreen.current.available ? 38 : 0)
```
DTCurrie: Maybe derive this in the `fullscreen` plugin, then we can just use it below as `fullscreen.baseOffset` or something?

src/lib/hooks/useFullscreen.svelte.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const provideFullscreen = () => {
+	const context = createFullscreen()
+	setContext<Context>(FULLSCREEN_CONTEXT_KEY, context)
+	return context
+}
+
+export const useFullscreen = () => {
+	return getContext<Context>(FULLSCREEN_CONTEXT_KEY)
+}
```
DTCurrie: we can probably simplify this a bit.

```suggestion
import { getContext, setContext } from 'svelte'

export const FULLSCREEN_CONTEXT_KEY = Symbol('fullscreen')

export const provideFullscreen = () => {
    const available = $state(false)
    const active = $state(false)

	return setContext(FULLSCREEN_CONTEXT_KEY, {
		get available() {
			return available
		},
		get active() {
			return active
		},
	})
}

export const useFullscreen = () => {
	return getContext<ReturnType<typeof provideFullscreen>>(FULLSCREEN_CONTEXT_KEY)
}
```

src/lib/hooks/useFullscreen.svelte.ts
DTCurrie, line 1: this hook should move to the `Fullscreen` plugin module directory.

src/lib/hooks/useFullscreen.svelte.ts
mrloureed: - Flattened the context as suggested (current, defaults(), createFullscreen gone; type inferred via ReturnType).
- Added setters — the getter-only version throws: Fullscreen.svelte assigns available/active and FullscreenButton binds active.
- let instead of const for the $state primitives, since they update by reassignment.
- Kept baseOffset (your other comment) as a $derived on the context.
- Moved the hook into lib/plugins/Fullscreen/ to colocate the feature.

One flag: this now diverges from the current wrapper pattern used by settings/environment

DTCurrie: cool with the departure, I think this is a pattern we will want to move forward with. we can discuss after we get the last of the plugins in and update everything to whichever pattern we choose.

DTCurrie (approved):

src/lib/plugins/Fullscreen/FullscreenButton.svelte
```diff
@@ -0,0 +1,69 @@
...
+	})
+</script>
+
+<svelte:window
+	onkeydown={(event) => {
+		if (event.key === 'Escape' && fullscreen) {
+			fullscreen = false
+		}
+	}}
+/>
```
micheal-parks: I'd recommend removing this. Escape is already used in multiple places. I don't think people will be toggling fullscreen enough for this to be a valuable hotkey.

PR #833, opened 2026-07-09
Fix transform controls & fragment editing

micheal-parks:
### Overview

This PR fixes an issue that silently popped up during instanced rendering updates and was exposed by https://github.com/viamrobotics/visualization/pull/821. The selected transform controls anchor themselves to Object3Ds, which no longer exist with instanced rendering. However, before the cleanup PR, ghost object3Ds were still being created via the `<Frame>` component that the selected transform controls could anchor to. The cleanup PR removed those and broke transform controls. This PR switches these controls over to use `WorldMatrix` like everything else, fixing the issue.

When fixing this, I discovered another fragment editing bug - fixed here - that recently became apparent due to vino2 config changes. Fragment tagging was incomplete in some scenarios when reading from the robot config, causing transform changes to not trigger the isDirty flag.

More details in comments.

### DIFF
--- a/src/lib/hooks/useFragmentInfo.svelte.ts
+++ b/src/lib/hooks/useFragmentInfo.svelte.ts
@@ -47,6 +47,33 @@ export const useFragmentInfo = (): FragmentInfoContext => {
 	return getContext<FragmentInfoContext>(key)
 }
 
+/**
+ * Extract the component names a `fragment_mods` list targets, from dot-notation
+ * mod paths like `components.<name>.frame` or `components.<name>.attributes.*`.
+ * A component only appears here if a fragment provides it, so its presence maps
+ * the component to that mod's fragment.
+ */
+const moddedComponentNames = (mods: unknown[]): string[] => {
+	const names: string[] = []
+	for (const mod of mods) {
+		if (!mod || typeof mod !== 'object') {
+			continue
+		}
+		for (const opValue of Object.values(mod as Record<string, unknown>)) {
+			if (!opValue || typeof opValue !== 'object') {
+				continue
+			}
+			for (const path of Object.keys(opValue as Record<string, unknown>)) {
+				const match = /^components\.([^.]+)/.exec(path)
+				if (match) {
+					names.push(match[1])
+				}
+			}
+		}
+	}
+	return names
+}
+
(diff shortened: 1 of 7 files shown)


mattmacf98 (approved):

src/lib/hooks/useFragmentInfo.svelte.ts
```diff
@@ -47,6 +47,33 @@ export const useFragmentInfo = (): FragmentInfoContext => {
...
+			for (const path of Object.keys(opValue as Record<string, unknown>)) {
+				const match = /^components\.([^.]+)/.exec(path)
+				if (match) {
+					names.push(match[1])
+				}
+			}
+		}
+	}
+	return names
+}
```
DTCurrie, line 75: nit: code convention/style question, not related to these actual changes except they are a good example. 

We oftentimes inline single-statement `if` conditions. Generally, I find it more readable, but I can see cases where it _could_ make them harder to read, especially when nested inside loops like the one above. 

Do we think this is more or less readable?


```suggestion
const moddedComponentNames = (mods: unknown[]): string[] => {
	const names: string[] = []
	for (const mod of mods) {
		if (!mod || typeof mod !== 'object') continue
		for (const opValue of Object.values(mod as Record<string, unknown>)) {
			if (!opValue || typeof opValue !== 'object') continue
			for (const path of Object.keys(opValue as Record<string, unknown>)) {
				const match = /^components\.([^.]+)/.exec(path)
				if (match) names.push(match[1])
			}
		}
	}
	return names
}
```

DTCurrie (approved):

src/lib/hooks/useFragmentInfo.svelte.ts
micheal-parks: I'm naturally very biased towards one statement per line, which is why a lot of my stuff gets written this way. Two statements per line seems arbitrary to me intuitively. It's sort of in the same bucket as `const a = 1; const b = 2;` being on the same IMO. I'm not opposed to the alternative style though if y'all feel differently, I'd just recommend we make it an auto-formatting rule rather than pointing it out in nits

DTCurrie: yeah that is why I am asking, I am wondering if we set this up as a linter/formatter thing

PR #838, opened 2026-07-10
Add motion plan JSON parser (parsePlan)

sucrammal:
The Motion Plan Replayer needs raw plan JSON turned into a typed, validated structure before anything can be rendered. Adds parsePlan, which parses a plan's concatenated JSON objects into a ParsedPlan (frames, parents, trajectory, goals) with zod validation and a typed PlanParseError.

## Changes
- Frontend: Adds parse-plan.ts — zod schemas + brace-matching splitter for multi-object plan JSON; exports parsePlan, ParsedPlan, RawFrame, PlanParseError
- Tests: Adds parse-plan.spec.ts covering valid plans, malformed JSON, and missing fields

## Stack
- PR #838 
- PR #839 
- PR #840 
- PR #841 
- PR #842
- PR #843
- PR #844

## Links
- [APP-9316](https://viam.atlassian.net/browse/APP-9316)

## Testing
- Run npx vitest run src/lib/plugins/MotionPlanReplayer/__tests__/parse-plan.spec.ts
- Confirm valid multi-object plan JSON parses into a ParsedPlan; malformed input throws PlanParseError with a readable message

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

[APP-9316]: https://viam.atlassian.net/browse/APP-9316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #839, opened 2026-07-10
Add frame-descriptor derivation for motion plans

sucrammal:
Rendering a plan step requires each frame resolved to a concrete transform, but a ParsedPlan only carries raw frame/orientation/geometry JSON. Adds buildFrameDescriptors, converting a ParsedPlan into static and jointed FrameDescriptors -> geometry, orientation conversion (quaternion/euler/orientation-vector), joint-index mapping, and end-effector reparenting.

## Changes
- Frontend: Adds build-frame-descriptors.ts (StaticFrameDescriptor / JointFrameDescriptor / FrameDescriptor) and plan-uuid.ts (stable 16-byte ids for transforms)
- Tests: Adds build-frame-descriptors.spec.ts covering static vs jointed frames, all three orientation encodings, geometry translation, model-frame parenting, and - OV/end-effector remapping

## Stack
- PR #838 
- PR #839 
- PR #840 
- PR #841 
- PR #842
- PR #843
- PR #844

##  Links

## Testing
- Run npx vitest run src/lib/plugins/MotionPlanReplayer/__tests__/build-frame-descriptors.spec.ts
- Confirm descriptors carry the correct parent, pose, and joint indices for each fixture

---

### DIFF
--- a/src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
+++ b/src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
@@ -1,0 +1,334 @@
+import { Euler, MathUtils, Quaternion, Vector3 } from 'three'
+
+import {
+	Capsule,
+	Geometry,
+	Pose,
+	RectangularPrism,
+	Sphere,
+	Vector3 as ViamVector3,
+} from '$lib/buf/common/v1/common_pb'
+import { OrientationVector } from '$lib/three/OrientationVector'
+import { quaternionToPose } from '$lib/transform'
+
+import type { ParsedPlan } from './parse-plan'
+
+import { planUuid } from './plan-uuid'
+
+/** A rigid link with a fixed local transform — no joint involved. */
+export interface StaticFrameDescriptor {
+	kind: 'static'
+	name: string
+	parent: string
+	localPose: Pose
+	geometry: Geometry | null
+	uuid: Uint8Array<ArrayBuffer>
+}
+
+/** A revolute joint frame. Its pose at each step is a pure rotation around `axis` by the trajectory angle. */
+export interface JointFrameDescriptor {
+	kind: 'joint'
+	name: string
+	parent: string
+	axis: { X: number; Y: number; Z: number }
+	componentName: string
+	jointIndex: number
+	uuid: Uint8Array<ArrayBuffer>
+}
+
+export type FrameDescriptor = StaticFrameDescriptor | JointFrameDescriptor
+
+// Shared scratch objects — safe in single-threaded JS
+const tmpQ = new Quaternion()
+const tmpQFrame = new Quaternion()
+const tmpQGeo = new Quaternion()
+const tmpQInv = new Quaternion()
+const tmpQLocal = new Quaternion()
+const tmpE = new Euler()
+const tmpV = new Vector3()
+const tmpOv = new OrientationVector()
+
+type QuatJson = { W: number; X: number; Y: number; Z: number }
+type EulerJson = { roll: number; pitch: number; yaw: number }
+type OvJson = { x: number; y: number; z: number; th: number }
+type OrientJson =
+	| { type: 'quaternion'; value: QuatJson }
+	| { type: 'euler_angles'; value: EulerJson }
+	| { type: 'ov_degrees'; value: OvJson }
+	| { type: 'ov_radians'; value: OvJson }
+	| undefined
+type Vec3Json = { X: number; Y: number; Z: number } | undefined
+
+const hasOrientJson = (orientation: OrientJson): orientation is NonNullable<OrientJson> =>
+	orientation?.type === 'quaternion' ||
+	orientation?.type === 'euler_angles' ||
+	orientation?.type === 'ov_degrees' ||
+	orientation?.type === 'ov_radians'
+
+/** Write orientation JSON into `out`. RDK euler_angles use Tait–Bryan Z-Y′-X″ (ZYX). */
+const quatFromJson = (orientation: OrientJson, out: Quaternion): Quaternion => {
+	if (orientation?.type === 'quaternion' && orientation.value) {
+		const v = orientation.value
+		// Three.js Quaternion order: (x, y, z, w) — W is LAST
+		return out.set(v.X, v.Y, v.Z, v.W)
+	}
+	if (orientation?.type === 'euler_angles' && orientation.value) {
+		const v = orientation.value
+		tmpE.set(v.roll, v.pitch, v.yaw, 'ZYX')
+		return out.setFromEuler(tmpE)
+	}
+	if (orientation?.type === 'ov_radians' && orientation.value) {
+		const v = orientation.value
+		return tmpOv.set(v.x, v.y, v.z, v.th).toQuaternion(out)
+	}
+	if (orientation?.type === 'ov_degrees' && orientation.value) {
+		const v = orientation.value
+		const th = MathUtils.degToRad(v.th ?? 0)
+		return tmpOv.set(v.x, v.y, v.z, th).toQuaternion(out)
+	}
+	return out.set(0, 0, 0, 1)
+}
+
+const poseFromFrame = (translation: Vec3Json, orientation: OrientJson): Pose => {
+	const pose = new Pose({
+		x: translation?.X ?? 0,
+		y: translation?.Y ?? 0,
+		z: translation?.Z ?? 0,
+	})
+	quaternionToPose(quatFromJson(orientation, tmpQ), pose)
+	return pose
+}
+
+type FramePoseJson = { translation?: Vec3Json; orientation?: OrientJson }
+
+/**
+ * Convert a geometry center pose from parent-frame coordinates into the link
+ * frame's local coordinates.
+ *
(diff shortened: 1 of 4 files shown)


src/lib/plugins/MotionPlanReplayer/plan-uuid.ts
```diff
@@ -0,0 +1,7 @@
+import { UuidTool } from 'uuid-tool'
+
+export const planUuid = (): Uint8Array<ArrayBuffer> => {
+	const bytes = new Uint8Array(new ArrayBuffer(16))
+	bytes.set(UuidTool.toBytes(crypto.randomUUID()))
```
mattmacf98: UUID tool already returns 16 bytes https://github.com/domske/uuid-tool/blob/3df4169458777941e04f58edb196f037c016881f/src/uuid-tool.ts#L4-L11

so you don't need a wrapper function for this and can just call the uuid tool directly

```
Uint8Array.from(UuidTool.toBytes(crypto.randomUUID()))
```

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+	const pose = new Pose({
+		x: translation?.X ?? 0,
+		y: translation?.Y ?? 0,
+		z: translation?.Z ?? 0,
+	})
+	quaternionToPose(quatFromJson(orientation, tmpQ), pose)
+	return pose
+}
+
+type FramePoseJson = { translation?: Vec3Json; orientation?: OrientJson }
```
mattmacf98: nit: id keep all types in one place at top of file

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+type QuatJson = { W: number; X: number; Y: number; Z: number }
+type EulerJson = { roll: number; pitch: number; yaw: number }
+type OvJson = { x: number; y: number; z: number; th: number }
+type OrientJson =
+	| { type: 'quaternion'; value: QuatJson }
+	| { type: 'euler_angles'; value: EulerJson }
+	| { type: 'ov_degrees'; value: OvJson }
+	| { type: 'ov_radians'; value: OvJson }
+	| undefined
+type Vec3Json = { X: number; Y: number; Z: number } | undefined
```
mattmacf98: why do these types have a `| undefined`? if you want a value to be this type or undefined we should do it like

`const val: Vec3Json | undefined` 

instead of putting undefined into the type itself

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+					y: trans?.Y ?? 0,
+					z: trans?.Z ?? 0,
+				})
+				if (hasOrientJson(orient)) {
+					quaternionToPose(quatFromJson(orient, tmpQ), local)
+				}
+				return local
+			})()
+
+	const label = (g.Label ?? g.label ?? '') as string
```
mattmacf98: why are there a `label` and a `Label` field? is this because of some inconsistency with motion plan structures?

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+
+	if (type === 'sphere') {
```
mattmacf98: make these switches and make it error on the case where the value is not sphere, capsule nor box instead of silently falling back to box

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+		label,
+	})
+}
+
+export const buildFrameDescriptors = (plan: ParsedPlan): FrameDescriptor[] => {
+	const { frames, parents } = plan
+
+	// Pass 1: map component name → ordered joint frame names (from "model" frames).
+	// This gives us the index each joint occupies in the trajectory array.
+	const jointMap = new Map<string, string[]>()
```
mattmacf98: nit I would split these into helper functions to make it a little more digestable
```
const jointMap = buildJointMap(frames)
const modelTerminalMap = buildModelTerminalMap(frames, parents, jointMap)

return buildDescriptors(plan, jointMap, modelTerminalMap)
```

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
```diff
@@ -0,0 +1,334 @@
...
+		const links = model?.links as Array<{ id: string }> | undefined
+		const endEffectorId = primaryOutput ?? links?.at(-1)?.id
+		if (endEffectorId) {
+			modelTerminalMap.set(frameName, `${frameName}:${endEffectorId}`)
+			continue
+		}
+		const jointNames = jointMap.get(frameName)
+		const lastJoint = jointNames?.at(-1)
+		if (!lastJoint) continue
+		for (const [childFrame, parentName] of Object.entries(parents)) {
```
mattmacf98: some of these processes would be easier if we built the node tree ahead of time

you could build a set of nodes and there parent node refs and child node refs first

then thing like this where we get the child of the last joint are a simple check on that node prop and you could also get the jointMap from that too (since you just need to check the children nodes of the model you want joints for)

src/lib/plugins/MotionPlanReplayer/plan-uuid.ts
sucrammal: Deleted redundant helper!

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
sucrammal: No its not an inconsistency in the plan format. Just got mixed up with the way json.Marshal outputs lowercase or uppercase fields. Dropped the redundant lowercase case, my fault here.

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
sucrammal: First half makes complete sense. Implemented the child --> parent index. But, not applicable to jointMap since there are some other components (like camera and gripper) that we don't derive from the model frame (e.g. left-arm) in the same way.

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
sucrammal: Refactored!

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
sucrammal: Yep, moved the default case as error.

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
sucrammal: Yep! I thought it'd save redundancy but its much less readable

src/lib/plugins/MotionPlanReplayer/build-frame-descriptors.ts
mattmacf98: ah ok yeah so arm would have `[arm:joint1, arm:joint2, camera]` as children and then we wouldn't know which one is a joint or not?

we could keep that info in a node type and then you have a map of component name -> node for lookups `Map<string,Node>`

but node would have refs to it's parent and children nodes

```
e.x. an arm:joint-1
type Node {
 name: "arm:joint-1"
 parent: "arm"
 children: [],
 isJoint: true
}

e.x. arm
type Node {
 name: "arm"
 parent: "world"
 children: ["arm:joint-1", "arm:joint-2", "camera"],
 isJoint: false
}

e.x. camera
type Node {
 name: "camera"
 parent: "arm"
 children: [],
 isJoint: false
}
```

sucrammal: Slight correction: so joints aren't children of the **model** frame like the gripper or camera:
```
children of model frame 'left-arm':
  ['left-cam_origin', 'left-gripper_origin', 'obstacle-arm-left-cord_origin']
```
Instead, the joints are under a different attribute under the model frame's kinematic definition:
```
frames["left-arm"] = {
  "frame_type": "model",
  "frame": { "model": {
    "links":  ["base", "base_top", "upper_arm", "upper_forearm", "lower_forearm", "wrist_link", "gripper_mount"],
    "joints": ["waist", "shoulder", "elbow", "forearm_rot", "wrist", "gripper_rot"]   // ← array ORDER = trajectory column
  }}
}
```
These 2 things are separate and will nonetheless need to be manually stitched together if we implement this graph. 

Additionally, we'll still need to go back to the joint order when we build the jointMap. The trajectory has a numbered array of joint angles for left arm, that we opaquely map to the joint order (e.g. 0 --> waist, 1 --> shoulder). we would need to encode this on the node graph as well. 

I'll try sketching this out now to see if it simplifies!

PR #841, opened 2026-07-10
Add Motion Plan Replayer plugin shell (panel + upload + extraSource)

sucrammal:
The Motion Plan Replayer currently has no entry point.  Adds the plugin shell â a dashboard-mounted floating panel that uploads plan JSON, parses it into snapshots via the pipeline, and lists plans with ready/error/no-trajectory status.

## Changes
- Adds MotionPlanReplayer.svelte (root + provider), MotionPlanReplayerUI.svelte (floating panel, upload, plan list), plan-dropper.ts (upload adapter), and useMotionPlanReplayer.svelte.ts (plan-list + parsed-snapshot store); mounts <MotionPlanReplayer /> in +layout.svelte and exports it from the plugin barrel
- extraSource snippet receives addPlan so an embedding app can inject its own plan source

## Stack
- PR #838 
- PR #839 
- PR #840 
- PR #841 
- PR #842
- PR #843
- PR #844

## Links
- [APP-9316](https://viam.atlassian.net/browse/APP-9316)

## Testing
- pnpm dev → open the Motion Plan Replayer panel from the dashboard → upload a plan JSON → confirm new row appears with eye icon enabled, suggesting it is ready. Upload malformed JSON, motion plan standalone request/response JSON  → confirm an error toast and error status. No 3D rendering yet.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

[APP-9316]: https://viam.atlassian.net/browse/APP-9316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #842, opened 2026-07-10
ScrubbingUI to visualize parsed motion plan JSON

sucrammal:
Plans parsed by the pipeline weren't rendered. This adds 3D rendering and a timeline scrubber: selecting a plan spawns its snapshot entities under a plan entity, and the scrubber resolves each slider position to the matching snapshot.

## Changes
- Adds MotionPlanReplayerScrubber.svelte (play/pause, step, seek, exit) and extends useMotionPlanReplayer.svelte.ts with applyStep/setStep, currentStep/totalSteps, and planEntity spawn/teardown
- Renders spawned entities via reconcileSnapshotEntities, tints them, and groups them with the PartOfPlan relation for one-shot cleanup
- Temporary fix to bump toasts up when scrubber is active. [Here is a quick PR](https://github.com/viamrobotics/app/pull/12754) to adjust the underlying prime component to do component placements instead of custom css class in viz. However, tailwind versions in visualization need upgrading to app side version, and need a prime version bump (visualization is minors behind current prime release), both of which will require some plumbing. 

## Stack
- PR #838 
- PR #839 
- PR #840 
- PR #841 
- PR #842
- PR #843
- PR #844

## Links

## Testing
- pnpm dev --> upload a plan --> confirm geometry renders and the bottom scrubber appears -->  drag/play to step through and watch the arm move --> exit or remove the plan and confirm all its entities disappear

---


src/app.css
```diff
@@ -67,3 +67,8 @@ select {
 .motion-tools-table tbody th {
 	@apply border-light font-roboto-mono text-default h-[40px] gap-2 border px-1.5 text-center text-xs font-normal;
 }
+
+/* Push toasts above the motion plan scrubber bar (fixed bottom-4, ~44px tall). */
+body.has-scrubber [aria-label='Toasts'] {
```
mattmacf98: maybe ok for now but would be nice if the scrubber and toasts where layed out in a way that svelte components could take care of nudging the toasts up when there is a scrubber

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -44,17 +60,76 @@ export const provideMotionPlanReplayer = (initialPlans?: PlanEntry[]) => {
...
+		planEntity = undefined
+		entityMap = new Map()
+		currentStep = 0
 		activePlanIndex = null
 	}
 
+	const tagPartOfPlan = (entity: Entity): void => {
+		if (!planEntity) return
+		entity.add(planRelations.PartOfPlan(planEntity))
+		for (const child of world.query(relations.ChildOf(entity))) {
```
mattmacf98: if we are tagging all of the entities spawned from a snapshot in a loop on line 92-97, why do we also have to swim down the node heirarchy? wont the children also be spawned and also be hit by that loop?

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
```diff
@@ -44,17 +60,76 @@ export const provideMotionPlanReplayer = (initialPlans?: PlanEntry[]) => {
...
+		if (planEntity) {
+			for (const entity of world.query(planRelations.PartOfPlan(planEntity))) {
+				if (!entity.isAlive()) continue
+				if (entity.has(traits.ReferenceFrame)) continue
+				if (entity.has(traits.Color)) {
+					entity.set(traits.Color, PLAN_COLOR)
+				} else {
+					entity.add(traits.Color(PLAN_COLOR))
+				}
+				entity.set(traits.Opacity, PLAN_OPACITY)
```
mattmacf98: how can you be sure it will have opacity trait (i.e. are you sure you don't need to add)?

src/app.css
sucrammal: You brought up this issue during scope PR! Do you think I should look into [modifying the underlying prime component here](https://github.com/viamrobotics/app/pull/12754)? @mrloureed

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Agreed on removing. Here's a longer explanation since it took me a bit to convince myself:
1. result.spawned is only the entities newly created on a scrubber step. It isn't one per spawned entity since we do diffing between steps.
2. if we have a model drawing, drawModel only spawns a root entity plus one ChildOf entity per GLTF asset and only returns the loop. Without this looping, we wouldn't be able to render these children. 
3. But, we don't have support for model drawings. So we're safe here.

src/lib/plugins/MotionPlanReplayer/useMotionPlanReplayer.svelte.ts
sucrammal: Done. Clearly didn't mirror the same as I did with color.

src/app.css
micheal-parks: This is fragile, and I don't think that any tool in the app should have the ability to influence toast location. I think instead we should rethink where to put this tool's UI if it's fighting visibility with toasts

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerScrubber.svelte
```diff
@@ -0,0 +1,176 @@
...
+	$effect(() => {
+		document.body.classList.toggle('has-scrubber', ctx.totalSteps > 0)
+		return () => document.body.classList.remove('has-scrubber')
+	})
+</script>
+
+<Portal id="dom">
+	{#if ctx.totalSteps > 0}
+		<div
+			class="pointer-events-auto fixed bottom-4 left-1/2 z-10000 flex w-[min(640px,calc(100vw-2rem))] -translate-x-1/2 items-center gap-3 rounded border border-[#666] bg-[#666] px-3 py-2 text-xs text-white"
```
micheal-parks: `z-10000` is pretty high, can we make this closer to the values of other floating UI?

src/app.css
sucrammal: @micheal-parks  Just aligned with Matthew: for now, we'll push the scrubber upward right below the dashboard buttons, and revert this fragile CSS hack. Will wait until Jason gets to this thread to give a proper suggestion or redesign if we want to fully rehome the scrubber bar.

micheal-parks: sounds good, thank you!

src/lib/plugins/MotionPlanReplayer/MotionPlanReplayerScrubber.svelte
sucrammal: Done!

PR #843, opened 2026-07-10
Persist per-frame display config + fix axes-helper toggle

sucrammal:
## Changes
Add persistent per-frame display config to the Motion Plan Replayer. 
- Per-frame color, opacity, visibility, and axes edits made via the scene Details panel and tree now persist across scrubbing instead of being reset each step. 
- Also fixes the Details panel "show axes helper" toggle, which previously had no effect because the batched axes renderer never reacted to the trait being added or removed at runtime.

## Stack
- PR #838 
- PR #839 
- PR #840 
- PR #841 
- PR #842
- PR #843
- PR #844

## Links

## Testing
- pnpm dev → open the Motion Plan Replayer panel from the dashboard, upload and select a plan. Click on various frames – obstacles, arm frames – and toggle the axes helpers on the details panel. Confirm that the axes helpers actually appear and disappear. 


mattmacf98 (approved):

src/lib/components/Entities/AxesHelpers.svelte
```diff
@@ -39,7 +39,7 @@
 		for (const entity of dirty) {
 			const worldMatrix = entity.get(traits.WorldMatrix)
 
-			if (entity.isAlive() && worldMatrix) {
+			if (entity.isAlive() && worldMatrix && entity.has(traits.ShowAxesHelper)) {
```
micheal-parks, line 42: I don't think this is the correct fix, I think that instead the toggle should be removing / adding the trait from the entity

sucrammal: We already do remove and add the trait from the entity via AxesHelperDetails.handleToggle, and the motion plan replayer's applyStep does the same. But, AxesHelpers.svelte doesn't subcribe to the ShowAxesHelper trait. On main it only listens for WorldMatrix/Center/InheritedInvisible + flush() here doesn't decide from the entity at all. 

TLDR it renders the axes helpers for whatever had the trait on at mount, then it would be deaf. Is there any other way you think we could fix this?

micheal-parks: Ah you know what, i totally misread this - you're right, this is a good fix, sorry about the confusion!

DTCurrie: so, this had e2e tests via the world state store service that _were_ working, so at some point something broke this.

micheal-parks: very likely that me switching axes to instanced rendering broke this

